PR #1, opened 2020-11-09
Upload Apache license and dco.txt

gilles-peskine-arm:
Upload the Apache license and indicate that all content is under this license.

Upload `dco.txt` (developer certificate of origin) so that we can set up an automated DCO check. Create a minimal `CONTRIBUTING.md`.

Part of https://github.com/ARMmbed/mbedtls/issues/3852.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,17 @@
+Contributing
+============
+
+We gratefully accept contributions from the community.
+
+As with any open source project, contributions will be reviewed by the project team and community and may need some modifications to be accepted.
+
+Making a Contribution
+---------------------
+
+For quick merging, the contribution should be short, and concentrated on a single topic. The larger the contribution is, the longer it would take to review it and merge it.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,17 @@
+Contributing
```
daverodgman, line 1: Shouldn't this include the Apache license header?

gilles-peskine-arm: Why?

I don't think it's reasonable to include a license header in documentation files. If we want every page to include a reference to the license, it should be injected during the build of the page.

daverodgman: I guess it seems inconsistent that we do this for code but not docs. I do agree that the license makes it clear that "unless otherwise indicated, all the content of this repository is distributed under the Apache License 2.0". So I'm fine with leaving it out.

daverodgman (approved):

PR #6, opened 2021-03-10
Create a directory for development tools

gilles-peskine-arm:


tools/README.md
```diff
@@ -0,0 +1,7 @@
+This directory contains various tools that are only of interest to
+developers of Mbed TLS.
```
chris-jones-arm, line 2: Might be worth explicitly stating that these tools could be of useful for anyone interested in developing mbedTLS. I think at the moment it comes off as quite closed that these tools are *only* of interest to developers and that they are for internal use only.

I think I would prefer something like:

```suggestion
This directory contains various tools that could be helpful to anyone
interested in developing Mbed TLS.
```

gilles-peskine-arm: I don't see a difference between “developers of” and “anyone interested in developing”. But I think your point is that this seemed to exclude occasional contributors, right? I've tweaked the wording to avoid this interpretation. The main point is that this is not a place for tools that are useful to users of Mbed TLS.

chris-jones-arm: Yes thank you that is what I was trying to say.

chris-jones-arm (approved):
Thanks for tweaking the wording, I'm happy with this now.

PR #12, opened 2021-04-30
New script to run a subset of unit tests based on their description

gilles-peskine-arm:
Run a subset of the unit tests in a test suite. The test cases are filtered by description. Run `mbedtls-run-tests --help` for a usage summary. You must build the test suite first.

For example, to run only the RSA tests of `test_suite_psa_crypto`:
```
make
mbedtls-run-tests -s RSA tests/test_suite_psa_crypto
```

And to run them in a debugger:
```
make
mbedtls-run-tests -c gdb -c --args -s RSA tests/test_suite_psa_crypto
```

I used this script once and it didn't break. Use at your own risk. Improvements welcome.

### DIFF
--- a/tools/bin/mbedtls-run-tests
+++ b/tools/bin/mbedtls-run-tests
@@ -74,6 +74,10 @@ def filter_test_cases(all_datax, tcf, temp_datax):
 def run_exe(keep_temp, precommand, exe, extra_options, all_datax, tcf):
     """Run one Mbed TLS test suites based on the specified options.
 
+    Return the subprocess's exit status (should be 0 for success, 1 for
+    a test failure, 2 on operational error), or a negative code if it was
+    killed by a signal.
+
     See `main` for what options are valid.
     """
     directory = os.path.dirname(exe)
@@ -121,8 +125,10 @@ def run(options):
         status = run_exe(options.keep,
                          options.command, exe, extra_options,
                          data_file, tcf)
-        if status > global_status:
-            global_status = status
+        if status > 0: # test failure or operational error
+            global_status = max(global_status, status)
+        elif status < 0: # killed by a signal
+            global_status = max(global_status, 120)
     return global_status
 
 def main():
--- a/tools/bin/mbedtls-run-tests
+++ b/tools/bin/mbedtls-run-tests
@@ -2,7 +2,7 @@
 
 """Run Mbed TLS unit tests.
 
-By default, run all test casess. If at least one of --prefix, --regex or
+By default, run all test cases. If at least one of --prefix, --regex or
 --substring is specified, only run the test cases that match any of these
 options.
 """
@@ -29,7 +29,7 @@ class TestCaseFilter:
             include += '|' + re.escape(string)
         for regex in options.regex:
             include += '|.*' + regex
-        if include: # no filter specified
+        if include:
             include = include[1:] # remove the leading '|'
         self.include = re.compile(include.encode())
         if options.exclude:
--- a/tools/bin/mbedtls-run-tests
+++ b/tools/bin/mbedtls-run-tests
@@ -1,0 +1,159 @@
+#!/usr/bin/env python3
+
+"""Run Mbed TLS unit tests.
+
+By default, run all test casess. If at least one of --prefix, --regex or
+--substring is specified, only run the test cases that match any of these
+options.
+"""
+
+import argparse
+import os
+import re
+import subprocess
+import tempfile
+
+
+class TestCaseFilter:
+    """Test case filter."""
+
+    def __init__(self, options):
+        """Set up a test case filter.
+
+        See `main` for what options are valid.
+        """
+        include = ''
+        for string in options.substring:
+            include += '|.*' + re.escape(string)
+        for string in options.prefix:
+            include += '|' + re.escape(string)
+        for regex in options.regex:
+            include += '|.*' + regex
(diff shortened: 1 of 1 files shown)


tools/bin/mbedtls-run-tests
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
+
+"""Run Mbed TLS unit tests.
+
+By default, run all test casess. If at least one of --prefix, --regex or
```
daverodgman, line 5: s/casess/cases

tools/bin/mbedtls-run-tests
```diff
@@ -0,0 +1,159 @@
...
+    """Test case filter."""
+
+    def __init__(self, options):
+        """Set up a test case filter.
+
+        See `main` for what options are valid.
+        """
+        include = ''
+        for string in options.substring:
+            include += '|.*' + re.escape(string)
```
daverodgman, line 27: probably safer to put round brackets around string - I think as-is, this would break if string contains "|"?

tools/bin/mbedtls-run-tests
```diff
@@ -0,0 +1,159 @@
...
+        for string in options.prefix:
+            include += '|' + re.escape(string)
+        for regex in options.regex:
+            include += '|.*' + regex
+        if include: # no filter specified
```
daverodgman, line 32: comment is slightly confusing - the condition is for when a filter *is* specified

tools/bin/mbedtls-run-tests
```diff
@@ -0,0 +1,159 @@
...
+            include = include[1:] # remove the leading '|'
+        self.include = re.compile(include.encode())
```
daverodgman, line 34: Is calling encode() needed here? I'm not sure why it would be.

tools/bin/mbedtls-run-tests
```diff
@@ -0,0 +1,159 @@
...
+    """Filter test cases.
+
+    Filter test cases from datax_file based on their description according
+    to the filter tcf. Write the selected test cases to temp_datax.
+    """
+    in_stanza = False
+    in_match = False
+    for line in open(all_datax, 'rb'):
+        was_in_stanza = in_stanza
+        in_stanza = not not line.strip()
```
daverodgman, line 65: The not not here is not needed

tools/bin/mbedtls-run-tests
```diff
@@ -0,0 +1,159 @@
...
+
+def filter_test_cases(all_datax, tcf, temp_datax):
+    """Filter test cases.
+
+    Filter test cases from datax_file based on their description according
+    to the filter tcf. Write the selected test cases to temp_datax.
+    """
+    in_stanza = False
+    in_match = False
+    for line in open(all_datax, 'rb'):
```
daverodgman, line 63: Possibly nicer to put all this in a "with open ... " block and then the temp_datax.flush() at the end is redundant

tools/bin/mbedtls-run-tests
```diff
@@ -0,0 +1,159 @@
...
+        extra_options.append('-v')
+    tcf = TestCaseFilter(options)
+    # Make sure we can find all the data files before we start running tests.
+    data_files = [find_data_file(exe) for exe in options.exes]
+    global_status = 0
+    for exe, data_file in zip(options.exes, data_files):
+        status = run_exe(options.keep,
+                         options.command, exe, extra_options,
+                         data_file, tcf)
+        if status > global_status:
```
daverodgman, line 124: Negative return codes (meaning the process was terminated by a signal) are possible - should probably catch this too (e.g. status = abs(run_exe ...)

daverodgman (changes requested):

tools/bin/mbedtls-run-tests
gilles-peskine-arm: Why would parentheses be useful? `re.escape` escapes special characters — and if it didn't parentheses wouldn't help.

tools/bin/mbedtls-run-tests
gilles-peskine-arm: `include.encode()` is a `bytes`, so `self.include` is a regex matching bytes, which is necessary because `filter_test_cases` opens the file in binary, which avoids barfing on non-ASCII characters if the ambient locale doesn't match the file's locale (at least with older Python versions that don't assume UTF-8).

Types would make this clearer but I haven't gotten into the habit of using types for low-maintenance productivity scripts. I think we should keep the tools in `mbedtls-docs` low-maintenance and not require things like mypy and pylint. If we decide the tool is important enough to warrant that, we should move it to the mbedtls repo.

tools/bin/mbedtls-run-tests
gilles-peskine-arm: Indeed, but I prefer to have a proper boolean variable.

tools/bin/mbedtls-run-tests
gilles-peskine-arm: `temp_datax` is already coming from the `with tempfile.NamedTemporaryFile` in the caller. How would using a `with open` for `all_datax` help?

tools/bin/mbedtls-run-tests
daverodgman: Good point, I missed that re.escape would deal with it

daverodgman (approved):

PR #13, opened 2021-06-03
mbedtls-prepare-build: update for test/src/drivers and add --cross

gilles-peskine-arm:
* Update for recent-ish 2.x and 3.0 versions with an extra source directory `test/src/drivers`.
* New option `--cross` to automate https://github.com/ARMmbed/mbedtls/issues/3795#issuecomment-714591143

I haven't yet done any work to handle the generation of configuration-independent files. Run `make generated_files` before running `mbedtls-prepare-build`.


gabor-mezei-arm (approved):
Looks good to me

PR #14, opened 2021-06-30
Script to find pull requests without a changelog entry

gilles-peskine-arm:


tools/bin/mbedtls-check-missing-changelog
```diff
@@ -0,0 +1,39 @@
...
+Find Mbed TLS pull requests in the specified range of revisions that seem
+to be missing a changelog entry.
+
+This script uses heuristics:
+* Pull requests are merge commits on the first-parent-path in the specified
+  revision range.
+* Pull requests that modify ChangeLog or ChangeLog.d in any way are assumed
+  to be ok.
+* Pull requests whose last commit before merging has an @arm.com author
+  are assumed to be team PR where we put a changelog entry if necessary.
```
gilles-peskine-arm, line 16: Nowadays I don't think that's relevant anymore. We do tend to forget changelogs even in team PRs, not significantly less so than in third-party PRs.

tools/bin/mbedtls-check-missing-changelog
```diff
@@ -0,0 +1,39 @@
...
+  help >&2
+  exit 120
+fi
+if [[ "$1" == '--help' ]]; then
+  help
+  exit
+fi
+
+revision_range=$1
+merge_commits=$(git log --merges --first-parent --format=%H $revision_range)
```
gilles-peskine-arm, line 30: Now that we always merge through GitHub, we could use a more sensitive search to only catch PR merges and not merges from inside PR.

tools/bin/mbedtls-check-missing-changelog
```diff
@@ -0,0 +1,39 @@
...
+for c in $merge_commits; do
+  if [[ $(git log -n1 --format=%ae $c\^2) == *@arm.com ]]; then
+    continue
+  fi
+  if git diff --name-only $c~1 $c | grep -q ChangeLog; then
+    continue
+  fi
+  git show --oneline $c | grep .
```
gilles-peskine-arm, line 38: I don't remember why I thought it useful to show the diff here. For a merge commit, that's a representation of conflicts, which is not particularly useful. Maybe I _didn't_ think it useful to show the diff, but just assumed that there would be no merge conflicts.

yanesca (approved):
LGTM.

PR #15, opened 2021-07-09
Migrate security advisories

davidhorstmann-arm:
This PR contains markdown source for the security advisories, taken from https://tls.mbed.org/tech-updates/security-advisories, scraped and converted to markdown by python script. After creating this, I noticed @yanesca's comment that the site already exists in an original markdown form somewhere, so this PR might be redundant.


yanesca:
> the site already exists in an original markdown form somewhere, so this PR might be redundant.

Not the whole site, just parts of it (the security advisories are such). We don't have access to the database and we need the devops team to do it for us. If they do it it, it is less work for us, but of course we can't know how long it takes for them to schedule and get the content for us. Now that you already have a script I don't think it makes sense to fetch the content from the database.

yanesca:
Usually when we use scripts to make changes, we add the script to the commit message if it is small enough. If it is a long script, we add it to the content and remove it in the last commit. The goal is to make reviews easier and preserve the script for future reference.

Could you please add your script to the PR too?

yanesca:
Fixes: https://github.com/ARMmbed/mbedtls/issues/3211

security-advisories/advisories/mbedtls-security-advisory-2015-01.md
```diff
@@ -0,0 +1,80 @@
...
+Security Advisories
+
+**Author:**
+Manuel Pégourié-Gonnard
+
+**Published:**
+Oct 5, 2015
+
+**Last updated:**
+Oct 8, 2015
```
yanesca, line 80: This is related to the structure of the old website, I don't think we want to keep this in the new documentation (I would expect readthedocs to generate something similar if we still decide that we need this). I think we should remove this part.

yanesca (changes requested):

yanesca (changes requested):
I have
- Reviewed the script
- Manually checked if all advisories are present
- Manually checked that the first and the last advisories are correct

Looks good to me!

Could you please add a commit removing the script? (This way it would still remain in the history, but won't be in the way when we hook this up into readthedocs.)

yanesca (approved):
LGTM.

PR #16, opened 2021-07-26
Add Sphinx configuration.

davidhorstmann-arm:
This adds the `conf.py` and `index.rst` files required by sphinx to build the documentation.

Signed-off-by: David Horstmann <david.horstmann@arm.com>

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -1,0 +1,52 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'Mbed TLS'
+copyright = '2021, The Mbed TLS Contributors'
+author = 'The Mbed TLS Contributors'
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = [
+]
+
+# Add any paths that contain templates here, relative to this directory.
+templates_path = ['_templates']
+
+# List of patterns, relative to source directory, that match files and
+# directories to ignore when looking for source files.
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
+
+# Add any paths that contain custom static files (such as style sheets) here,
+# relative to this directory. They are copied after the builtin static files,
+# so a file named "default.css" will overwrite the builtin "default.css".
+html_static_path = ['_static']
--- a/conf.py
+++ b/conf.py
@@ -35,7 +35,8 @@ templates_path = ['_templates']
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store',
+                    'README.md', 'CONTRIBUTING.md', 'tools/README.md']
 
 
 # -- Options for HTML output -------------------------------------------------
--- a/conf.py
+++ b/conf.py
@@ -36,7 +36,7 @@ templates_path = ['_templates']
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store',
-                    'README.md', 'CONTRIBUTING.md', 'tools/README.md']
+                    'README.md', 'tools/README.md']
 
 
 # -- Options for HTML output -------------------------------------------------
--- a/conf.py
+++ b/conf.py
@@ -18,7 +18,7 @@
 # -- Project information -----------------------------------------------------
 
 project = 'Mbed TLS'
-copyright = '2021, The Mbed TLS Contributors'
+copyright = 'The Mbed TLS Contributors'
 author = 'The Mbed TLS Contributors'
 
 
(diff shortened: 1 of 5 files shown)


davidhorstmann-arm:
Fixes ARMmbed/mbedtls#3643

index.rst
```diff
@@ -0,0 +1,23 @@
...
+===================================
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
+   reviews/review_guidelines.md
+   architecture/proposed/long-term-plans.md
+   security-advisories/security-advisories.md
+   maintainers.md
```
gilles-peskine-arm, line 16: Was this automatically added by `sphinx-quickstart`? It wasn't when I tried it with Sphinx 1.8.5. If not, please make that in a separate commit.

conf.py
```diff
@@ -0,0 +1,170 @@
...
+# You can specify multiple suffix as a list of string:
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = ['.rst', '.md']
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = 'MbedTLSDocs'
```
gilles-peskine-arm, line 49: Why this weird StudlyCapsifiedName? Shouldn't the project name be “Mbed TLS”?

conf.py
```diff
@@ -0,0 +1,170 @@
...
+copyright = '2021, MbedTLS Developers'
```
gilles-peskine-arm, line 50: Our copyright notices say “The Mbed TLS Contributors”.

conf.py
```diff
@@ -0,0 +1,170 @@
...
+author = 'MbedTLS Developers'
```
gilles-peskine-arm, line 51: It's spelled “Mbed TLS”, not “MbedTLS”. And this should probably be the same wording we use for copyright notices.

conf.py
gilles-peskine-arm, line 50: I think the copyright should be constructed from `authors`. I'm not sure we actually need a year here: we've removed years from the copyright notices in the source code, but I don't know what the legal implications are. When we had years, the copyright notices had 2006–&lt;year of last modification&gt;. If a year is needed, considering that this is for the whole documentation and it changes often, we could use the current year: `copyright = now.strftime('%Y') + ' ' + author`

conf.py
```diff
@@ -0,0 +1,170 @@
...
+# -- General configuration ------------------------------------------------
+
+# If your documentation needs a minimal Sphinx version, state it here.
+#
+# needs_sphinx = '1.0'
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = ['myst_parser']
```
gilles-peskine-arm, line 34: Please add `myst_parser` in a separate commit explaining what it's for. Also, since it doesn't seem to be distributed with sphinx itself, where to get it.

conf.py
```diff
@@ -0,0 +1,170 @@
...
+
+# Add any paths that contain templates here, relative to this directory.
+templates_path = ['_templates']
+
+# The suffix(es) of source filenames.
+# You can specify multiple suffix as a list of string:
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = ['.rst', '.md']
```
gilles-peskine-arm, line 43: `.md` wasn't put there automatically by `sphinx-quickstart`, was it? I guess this should be added in the same commit that adds `myst_parser`.

gilles-peskine-arm (changes requested):

index.rst
davidhorstmann-arm: Ack - this was added by me, so I'll do a separate commit.

conf.py
davidhorstmann-arm: The name of this project must be distinct from "Mbed TLS" as this will eventually be a subproject of the main Mbed TLS project containing the versioned documentation (since the docs are distributed across 2 separate repos).

I take your point on the CamelCase though - that's my overactive aversion to spaces manifesting itself. I'll fix this in the next one.

gilles-peskine-arm: `project` is a human-readable name, not an internal identifier. It should be spelled the way a human would spell it, with spaces where warranted.

conf.py
```diff
@@ -35,7 +35,8 @@
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store',
+                    'README.md', 'CONTRIBUTING.md', 'tools/README.md']
```
gilles-peskine-arm, line 39: Why is `CONTRIBUTING.md` excluded? Will it maybe be linked from a different place? It is useful for people who read the documentation to know how to contribute to it.

conf.py
```diff
@@ -0,0 +1,52 @@
+# Configuration file for the Sphinx documentation builder.
```
gilles-peskine-arm, line 1: I think creating a makefile to build the sphinx documentation is within the scope of this task. We're adding a way to build the documentation so we should know _how_ to build it. (Creating a CI for it would be a follow-up however.) The simple one generated by `sphinx-quickstart` would do.

index.rst
```diff
@@ -0,0 +1,23 @@
+.. Mbed TLS documentation master file, created by
+   sphinx-quickstart on Thu Jul 29 17:27:29 2021.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to Mbed TLS's documentation!
```
gilles-peskine-arm, line 6: Maybe it's just me but I don't like this title. “Welcome to” is just noise: the title should start with the most important part, which is “Mbed TLS”. “Mbed TLS's” is awkward, “Mbed TLS documentation” would work just fine. I think there should be another qualifier there however to say that this is the central place for documentation, as opposed, say, to the documentation of a specific version. How about
```suggestion
Mbed TLS documentation hub
```

index.rst
```diff
@@ -0,0 +1,23 @@
...
+====================================
+
+.. toctree::
+   :maxdepth: 2
```
gilles-peskine-arm, line 10: (likely non-actionable for now)  A depth of 2 looks about right but I don't think the current structure works. The home page shouldn't have direct links to sections of each subpage. But the subpages should be grouped by topic: review guidelines under contributing, security advisories under news or security, long-term plans and maintainers under project, ...

But it's a bit hard to anticipate the right structure when we have so little content. We can go with what we have for now and restructure it as we go.

conf.py
gilles-peskine-arm, line 1: (likely non-actionable for now) The build comes up with many warnings. They're cosmetic so no big deal. Eventually I think we should enforce a build without warnings but it's too early for that.

index.rst
```diff
@@ -0,0 +1,23 @@
...
+   :caption: Contents:
+
```
gilles-peskine-arm, line 12: Before this can become the home page of Mbed TLS on the web, this needs some important bits of information:

* What this site is (a place for documentation).
* What this site isn't (finished).
* Where to find what isn't on this site (github, trustedfirmware, old website).

gilles-peskine-arm (changes requested):
The history looks good now, thanks. The finished version needs a little polishing to be publishable.

PR #19, opened 2021-12-08
mbedtls-prepare-build: add support for coverage with lcov

gilles-peskine-arm:
I had this lying around.

### DIFF
--- a/tools/bin/mbedtls-prepare-build
+++ b/tools/bin/mbedtls-prepare-build
@@ -1185,6 +1185,25 @@ if [ -n "$$failures" ]; then echo; echo "Failures:$$failures"; false; fi
         self.add_clean('$(test_apps)')
         # TODO: test_psa_constant_names.py
 
+    COVERAGE_RULE = """
+        rm -rf Coverage
+        lcov --capture --initial --directory library -o files.info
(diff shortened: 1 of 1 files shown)


tools/bin/mbedtls-prepare-build
```diff
@@ -1185,6 +1185,25 @@ if [ -n "$$failures" ]; then echo; echo "Failures:$$failures"; false; fi
         self.add_clean('$(test_apps)')
         # TODO: test_psa_constant_names.py
 
+    COVERAGE_RULE = """
```
minosgalanakis, line 1188: Any particular reason we do not define it as a list of strings?

COVERAGE_RULE= ["command A",
                                "command B",
                                "command C" ]

Rather that a docstring which we proccess in an iterator?

minosgalanakis: LGTM

minosgalanakis:
LTGM

tools/bin/mbedtls-prepare-build
gilles-peskine-arm: No particular reason. I guess I felt that since it's a long succession of commands with no variables, it was more readable that way. Also it's copied from `Makefile` and makes it easier to reconcile any changes that way.

minosgalanakis (approved):

PR #22, opened 2022-01-05
Add newest security advisory

paul-elliott-arm:
## Description

Add the latest security advisory to the docs.

## Notes 

Although overview file says it is generated, I didn't see any method of generation, so just attempted to follow the pattern as best as possible.

### DIFF
--- a/security-advisories/security-advisories.md
+++ b/security-advisories/security-advisories.md
@@ -5,7 +5,13 @@
 This is a generated list of our released advisories. A more comprehensive view
 is available in the Security Center.
 
-28 advisories have been released.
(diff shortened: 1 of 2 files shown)


security-advisories/security-advisories.md
```diff
@@ -5,7 +5,13 @@
 This is a generated list of our released advisories. A more comprehensive view
 is available in the Security Center.
```
gilles-peskine-arm, line 6: This sentence was copied from the website where the list was dynamically generated.

gilles-peskine-arm (approved):

PR #23, opened 2022-02-24
Scripts to watch the evolution of generated files

gilles-peskine-arm:
Archive the evolution of generated files over a range of revisions.

### DIFF
--- a/tools/bin/mbedtls-trace-files.py
+++ b/tools/bin/mbedtls-trace-files.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python3
+
+"""Archive the contents of the specified files for the specified Git revisions.
+
+Run this script from a clean Git worktree.
+This script runs `make FILE` to generate the desired files.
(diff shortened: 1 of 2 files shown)


tools/bin/mbedtls-trace-files.py
```diff
@@ -0,0 +1,186 @@
...
+        `files`: list of files to archive.
+        """
+        subprocess.check_call(['git', 'checkout', revision])
+        # In case the make command doesn't have correct dependencies,
+        # make sure that it at least regenerates the targets. This isn't
+        # enough if intermediate files need to be updated as well, but
+        # it's the best we can do here. If make doesn't have correct
+        # dependencies, make sure to delete all relevant intermediate files
+        # in the run_before command.
+        for filename in files:
```
bjwtaylor: Does this block need to account for if the --build-dir argument is used and remove the files relative to this directory, rather than the current one?

PR #25, opened 2022-04-20
Add document describing review prioritisation

daverodgman:
Signed-off-by: Dave Rodgman <dave.rodgman@arm.com>

### DIFF
--- a/reviews/review-priorities.md
+++ b/reviews/review-priorities.md
@@ -1,0 +1,68 @@
+# PR Prioritisation
+
+The Mbed TLS core team prioritise PRs to help us spend our review capacity on the items that best help the project.
+
+When PRs are first triaged, a label is applied to help reviewers determine what to work on (and what not to work on). The different classifications are described below.
+
+Please note that if your PR falls into the lower priority categories, we are happy to increase its priority if needed. This might happen through:
+
+- Commenting on the PR to explain why there is an important use-case for it
+- Other users commenting to show wider demand for the PR
+- Attending the bi-weekly Tech Forum call to discuss the PR
+- Advocating for the PR on the project mailing list
+- Find others to review your PR
+
+Classifications are done through labels, e.g.: priority-high.
+
+Please note that these are the priorities used by the core team. External contributors are welcome to review PRs according to their own priorities, and such review activity will help highlight PRs which should get attention from the core team.
+
+
+### priority-high
+
+This classification is used for PRs which are typically one of:
+
+- Security (or other high-priority) bug-fixes
+- Items that the Mbed TLS team have planned for delivery in the current quarter
+
+These PRs should receive review promptly.
+
+
+### priority-medium
+
+This is for PRs which advance the core team's aims for Mbed TLS, but are not needed to meet delivery in the current quarter. These would typically address one of the following areas:
+
+* Security
+  * Non-urgent and lower-priority security improvements
+* Robustness
+  * Non-urgent bug-fixes
+  * Test improvements
+  * Documentation
+* PSA Crypto
+  * Provide a reference implementation for PSA Crypto
+  * Driver interface for secure elements & crypto processors
+* TLS
+  * Support latest TLS versions and features - especially client features important for embedded devices
+  * DTLS, QUIC
+* Embedded applications
+  * Improvements to code-size, memory footprint and performance on embedded devices
+
+
+### priority-low
+
+This is for PRs which are potentially useful or nice to have, but are unlikely to receive review soon. This might include PRs in the following areas:
+
+* New ciphers for which there is low demand
+* Minor performance enhancements
+* Enhancements that don't suit embedded applications
+* Code (especially assembly) for platforms where the team has little experience for review or maintenance, or limited ability to test
+  * Note: we currently have the capability to maintain & test aarch32, aarch64, x86 and x86-64 on Ubuntu, FreeBSD and Windows
+
+* Minor quality/test improvements, e.g. clean-up of test infrastructure or build scripts
+* Items where the team might not be able to afford the maintenance cost
+
+After at least one year, some items in this category may be closed if a clear need does not emerge with time. This is more likely to happen for larger and more complex PRs.
+
+
+### priority-scheduled
+
+This is an additional label used for larger PRs (typically priority-medium) which require significant team bandwidth to review. These items will not be reviewed immediately, but they will be scheduled in the project roadmap with time allocated for review.
--- a/reviews/review-priorities.md
+++ b/reviews/review-priorities.md
@@ -8,7 +8,7 @@ Please note that if your PR falls into the lower priority categories, we are hap
 
 - Commenting on the PR to explain why there is an important use-case for it
 - Other users commenting to show wider demand for the PR
-- Attending the bi-weekly Tech Forum call to discuss the PR
+- Attending the regular Tech Forum call to discuss the PR
 - Advocating for the PR on the project mailing list
 - Find others to review your PR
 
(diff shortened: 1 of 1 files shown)


reviews/review-priorities.md
```diff
@@ -0,0 +1,68 @@
+# PR Prioritisation
+
+The Mbed TLS core team prioritise PRs to help us spend our review capacity on the items that best help the project.
+
+When PRs are first triaged, a label is applied to help reviewers determine what to work on (and what not to work on). The different classifications are described below.
```
gilles-peskine-arm, line 5: What's the advantage of labels, compared to columns on a board?

daverodgman: As discussed earlier: there are a few ways to capture this kind of thing in GitHub.

Labels are an obvious choice because they are simple, easily visible to people viewing the PR, quick to set, etc. They also interact well with our existing boards - no other changes are required to our setup. They are easy for people who don't know the process well to observe and make sense of. It's easy to filter to find the high-priority items in any GitHub context (e.g., searching PRs, viewing a board).

You suggested capturing this via a column on the unified board, which I don't think works so well. The unified board tracks PR/issue _progress_, not priority. We'd need to add extra columns for the various priorities, for both review and (less critically) stalled columns - if done thoroughly it would add a lot of complexity to this board. Outside / casual contributors may not know to look at the board so might not follow the prioritisation. It also wouldn't be visible in the epics board, PR search results, etc.

The main argument against using labels for this seems to be that we have too many labels (57 now). This is very true and I support simplifying the set of labels. E.g., there are quite a few relating to the area affected (component: <X>, TLS 1.3/MPS, PSA, etc), which we could drop; also some obsolete ones relating to things like CLA, baremetal, etc. There are some that are hardly used (e.g., Information, notabug, regression, wontfix) which could go and just be replaced by an appropriate comment in the text of the PR/issue.

gilles-peskine-arm (approved):
I'm not personally convinced that labels are the way to go, but if we're going to use labels, I think this is a good policy.

reviews/review-priorities.md
gilles-peskine-arm: > This is very true and I support simplifying the set of labels.

The thing is, we all agree on that, but we can't agree about which labels to remove. Different labels are useful to different people.

Of the ones you're listed, the only ones I agree with are the obsolete ones, where we could make a list of labelled PR/issues once and for all and then remove the labels.

reviews/review-priorities.md
```diff
@@ -0,0 +1,68 @@
...
+  * Documentation
+* PSA Crypto
+  * Provide a reference implementation for PSA Crypto
+  * Driver interface for secure elements & crypto processors
+* TLS
+  * Support latest TLS versions and features - especially client features important for embedded devices
+  * DTLS, QUIC
+* Embedded applications
+  * Improvements to code-size, memory footprint and performance on embedded devices
+
```
tom-cosgrove-arm, line 48: There's no equivalent of the sentence "These PRs should receive review promptly" that the high-priority case has. Would it help to be explicit with something like the following?

`These PRs will receive review on a best-effort basis, but assistance from the community would be helpful`

(or whatever we actually want to say)

tom-cosgrove-arm: (note that we have an equivalent for low-priority of `unlikely to receive review soon`, so I think it would be helpful to be explicit as to what our promise is or isn't, rather than just leaving it implicitly somewhere between the high and low guarantees)

reviews/review-priorities.md
```diff
@@ -0,0 +1,68 @@
...
+
+### priority-medium
+
+This is for PRs which advance the core team's aims for Mbed TLS, but are not needed to meet delivery in the current quarter. These would typically address one of the following areas:
+
+* Security
+  * Non-urgent and lower-priority security improvements
+* Robustness
+  * Non-urgent bug-fixes
```
tom-cosgrove-arm, line 37: (very minor: is `bug-fix` normally hyphenated?)

reviews/review-priorities.md
```diff
@@ -0,0 +1,68 @@
...
+Classifications are done through labels, e.g.: priority-high.
+
+Please note that these are the priorities used by the core team. External contributors are welcome to review PRs according to their own priorities, and such review activity will help highlight PRs which should get attention from the core team.
+
+
+### priority-high
+
+This classification is used for PRs which are typically one of:
+
+- Security (or other high-priority) bug-fixes
```
tom-cosgrove-arm, line 24: (very minor: is `bug-fix` normally hyphenated?)

reviews/review-priorities.md
```diff
@@ -0,0 +1,68 @@
...
+
+The Mbed TLS core team prioritise PRs to help us spend our review capacity on the items that best help the project.
+
+When PRs are first triaged, a label is applied to help reviewers determine what to work on (and what not to work on). The different classifications are described below.
+
+Please note that if your PR falls into the lower priority categories, we are happy to increase its priority if needed. This might happen through:
+
+- Commenting on the PR to explain why there is an important use-case for it
+- Other users commenting to show wider demand for the PR
+- Attending the bi-weekly Tech Forum call to discuss the PR
```
tom-cosgrove-arm, line 11: (aside: bi-weekly has two meanings: twice a week or every two weeks)

```suggestion
- Attending the Tech Forum call (which is every two weeks) to discuss the PR
```

tom-cosgrove-arm (approved):

daverodgman:
Open question to reviewers: the top priority here includes items that are scheduled for the current quarter, i.e. could be delivered in up to three months time. Is that too broad, and do we need a level for things that are to be done as soon as possible (e.g., urgent bug-fixes)? We could always add something later so I'm inclined to stick to these options for now, WDYT?

tom-cosgrove-arm:
We can always add `priority-critical` at a later date if we find we need it, so let's just go with H M L for now

gilles-peskine-arm:
> the top priority here includes items that are scheduled for the current quarter, i.e. could be delivered in up to three months time. Is that too broad, and do we need a level for things that are to be done as soon as possible (e.g., urgent bug-fixes)? We could always add something later so I'm inclined to stick to these options for now, WDYT?

“We're definitely actively working on it” (“priority: high”) and “we absolutely want it in the next release” (which we aren't tracking in a systematic way, only through certain epics and from one special case, the `regression` label) are different concepts. If we're using labels for these, I'd prefer:

* `priority: high` is only for things that we want to _release_ ASAP. Not for something we want to _do_ quickly because it's a prerequisite of a prerequisite of a prerequisite of a delivery scheduled in six months.
* And then we introduce a `priority: commitment` for things we've committed to delivering, which are important but not urgent.

laumor01:
what about mirroring ARM commitments levels: 
Committed = High priority = Must Do (label assigned to issues/PRs that are required for a committed/diamond Epic)
Tentative = Medium Priority = Should Do (label assigned to issues/PRs that are required for a tentative/orange Epic)
Requested = Low Priority = Could Do (label assigned to issues/PRs that are required for a Requested/lemon Epic)

daverodgman:
> what about mirroring ARM commitments levels: Committed = High priority = Must Do (label assigned to issues/PRs that are required for a committed/diamond Epic) Tentative = Medium Priority = Should Do (label assigned to issues/PRs that are required for a tentative/orange Epic) Requested = Low Priority = Could Do (label assigned to issues/PRs that are required for a Requested/lemon Epic)

I think this would work for people within OSS/Arm, but less meaningful for people in the broader TF.org or general community. It obviously maps very well onto internal roadmap items, but it doesn't cover non-roadmap work like security fixes, tech debt, community review, CI work, releases, etc etc. The aim here is a broader Mbed TLS prioritisation process that is usable from our internal project management perspective, but also addresses community needs & priorities.

My thinking on this mapping was that all of our must and should do items would be priority-high. Within the team we would then highlight the diamond commitments in weekly calls etc, but don't need to capture the distinction on GitHub.

PR #32, opened 2022-07-15
Remove links to empty indices and search page

bensze01:
The indices would automatically populate with function definitions, which are not present in the unversioned documentation.
If we decide to manually create a traditional word-index, we can re-add the link to the index page.

The search page is empty when using readthedocs's default theme.
Search is instead handled through the sidebar.


AndrzejKurek (approved):
LGTM.

wernerlewis (approved):
LGTM

PR #35, opened 2022-07-26
Update mbedtls-security-advisory-2022-07.md

KloolK:
Removed some accidentally pasted text


bensze01 (approved):
Thanks for catching this! The PR looks good to me.

PR #36, opened 2022-08-02
Convert advisory list to auto-generated toctree

bensze01:
The list of advisories currently needs to be updated each time a new one is added.  
This PR implements auto-generation of the list.

It also resolves the warning emitted because there were no entries to the advisory pages in the toctree.  
This means the advisories show up in the sidebar, and in pdf builds of the documentation.


security-advisories/advisories/mbedtls-security-advisory-2017-01.md
```diff
@@ -8,7 +8,7 @@ for each possible attack.
 **Title** |  Remote plaintext recovery on use of CBC based ciphersuites
 ---|---
 **CVE** |  CVE-2018-0497
-**Date** |  25th July 2018
+**Date** |  25th July 2018 ( **Updated on 26th July 2018** )
```
wernerlewis, line 11: Unrelated to the changes made in this PR, but this advisory describes the same vulnerabilities as `2018-02`, with almost the same file content. From checking the wayback machine, it appears that the original `2017-01` was replaced at some point in 2018: https://web.archive.org/web/20180219050800/https://tls.mbed.org/tech-updates/security-advisories/mbedtls-security-advisory-2017-01

wernerlewis (approved):
LGTM, I checked each of the added "Updated on" dates, and these match those previously listed for each in `security-advisories.md`.

security-advisories/advisories/mbedtls-security-advisory-2017-01.md
gilles-peskine-arm: Indeed it looks like we made a mistake when uploading 2018-02.

bensze01: Thanks for catching this! Will file a separate PR to restore the original.

PR #38, opened 2022-08-02
Fix warnings about non-consecutive heading levels

bensze01:
Sphinx warns about non-consecutive headings and headings starting at level 2+ in any document during builds.  
It corrects these errors before rendering the webpages, so the output shouldn't be changed by this PR.


wernerlewis (approved):
LGTM

PR #39, opened 2022-08-05
Add knowledge base to documentation

bensze01:
This PR adds the knowledge base articles from the knowledge base repo to the documentation.

Old articles from the website that haven't been ported in the knowledge base repo have not been added.


kb/development/controlling_package_size.md
```diff
@@ -0,0 +1,168 @@
+# Mbed TLS over low-bandwidth, unreliable datagram networks
```
gilles-peskine-arm, line 1: This file was under `kb/how-to`.

kb/compiling-and-building/using-mbedtls-in-microsoft-visual-studio-2015.md
```diff
@@ -0,0 +1,519 @@
...
+
+## Use the solution file supplied with the Mbed TLS source code
+
+### Open the Visual Studio solution
+
+You now have this folder: My Documents\Visual Studio 2015\Projects\mbedtls-<version\>\visualc\VS2010.
+
+In this folder, you can find the file `mbedTLS.sln`. This is a Visual Studio `solution` file, which holds all the components Mbed TLS consists of and the rules to build the Mbed TLS library.
+
+<span class="images">![](https://mbed-tls-docs-images.s3.amazonaws.com/MbedTLS-Tutorial-VS2015-1.png)</span>
```
gilles-peskine-arm, line 30: The images in this file (and a few in other files) have absolute URLs which are broken, and the corresponding images are not in this pull request.

index.md
```diff
@@ -19,4 +19,5 @@ architecture/proposed/long-term-plans.md
 security-advisories/security-advisories.md
 maintainers.md
 CONTRIBUTING.md
+kb/index.md
```
gilles-peskine-arm, line 22: The paths to the old knowledge base were e.g. `https://tls.mbed.org/kb/how-to/add-a-random-generator` with no extension. Now the articles are at e.g. `https://tls.mbed.org/kb/how-to/add-a-random-generator.html`, i.e. with `.html` at the end. So existing links still don't work. But at least the articles are available.

gilles-peskine-arm (approved):
I made some spot checks, not a systematic review. This isn't perfect: there are still broken links. But it's a huge improvement: at least the articles are available and can be found from the home page. So I'm approving as is, and we can improve in follow-ups.

index.md
bensze01: We can handle this in two ways - the sphinx build corresponding to the old naming scheme is `dirhtml`, instead of the `html` we are currently using.

Readthedocs also supports adding redirections between the two schemes automatically.
So we can switch to the old scheme (and redirect <foo>.html to <foo>), or just switch on redirections from <foo> to <foo>.html.

Both options are fairly easy to achieve.

kb/development/controlling_package_size.md
gilles-peskine-arm: https://github.com/Mbed-TLS/mbedtls-docs/pull/41

index.md
gilles-peskine-arm: https://github.com/Mbed-TLS/mbedtls-docs/pull/40 removes the `.html` extension from the new site.

This is only part of the solution because the current redirection on `tls.mbed.org` redirects all pages to `https://www.trustedfirmware.org/projects/mbed-tls/`. We need to change this to a redirection that remembers the source and redirects under `https://mbed-tls.readthedocs.io/en/latest/` for most URLs.

PR #41, opened 2022-08-08
Move kb article to match old website layout

bensze01:
The old website listed `controlling_package_size` under `kb/how-to/`.


gilles-peskine-arm (approved):

gilles-peskine-arm:
I just happened to notice this one, but I didn't make systematic checks. Did you check if there were other misplaced files? Or are you waiting until everything is re-created to do a check on all old working URLs (which you'd get a list of… somehow)?

bensze01:
The KBs are all in the right place (the ones that were ported in the docs repo anyhow). I still need to move the security advisories to the correct places.

I'm using the list of files from yuto_static branch of the old website repo (you need to remove the .html extension to get the actual URL the files were served under).

If you check out the yuto_static branch from the old website repo you can use the following command to check for discrepencies:
```sh
diff --color=auto -u <(cd <mbedtls-docs repo>; find kb -type f | sed 's/\(\/index\)\?\.md/.html/' | sort) <(cd <old website repo, yuto_static branch>/htdocs && { find kb* -type f; } | sort)
```

PR #43, opened 2022-08-09
Move security advisories to match old website

bensze01:
Move the security advisories to match the path  they used to have on the old website.

I've added redirects from the old paths to the new paths on the PR version of the documentation.
This will need to be replicated for the live version when the PR is merged.


wernerlewis (approved):
LGTM

PR #46, opened 2022-08-30
Minor knowledge base updates for Mbed TLS 2.2x and 3.x

gilles-peskine-arm:
Several minor updates to the knowledge base, mostly for things that were true when written but stopped applying in some 2.x version or in 3.0.

* `config.pl` is now `config.py` (long overdue!)
* `config.h` is now `mbedtls_config.h` (but still acknowledge `config.h`: it's still there in 2.28 LTS)
* Update GitHub `ARMmbed` links to `Mbed-TLS`
* Acknowledge PSA crypto in a few places

I do not have completeness goals here. If I missed a `config.pl` somewhere, this pull request is incomplete. If something completely different needs updating, or if an article needs to have major updates, it's out of scope, and please request or make a follow-up. In particular:

* I didn't add a mention of `MBEDTLS_PSA_CRYPTO_CONFIG`, both because it's a bit too big for this update and because it's still officially an experimental feature.
* `/api` links are broken. We can't fix this until we have somewhere to point them to.
*

### DIFF
--- a/kb/compiling-and-building/compile-mbedtls-to-a-static-library-in-eclipse-cdt.md
+++ b/kb/compiling-and-building/compile-mbedtls-to-a-static-library-in-eclipse-cdt.md
@@ -23,7 +23,7 @@ This tutorial will show how to compile Mbed TLS to a static `.a ` library file i
 * Create your new C project by choosing  **File** > **New C Project**.
 * The **C Project** dialog pops up:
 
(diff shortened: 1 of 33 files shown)


tom-cosgrove-arm (approved):
Makes sense to get this in quickly and iterate on it

bensze01 (approved):
Looks good to me.

PR #47, opened 2022-09-07
Coding standards: loosen naming rules for static functions

gilles-peskine-arm:
Proposal to change coding standards: do not forbid the `mbedtls_` prefix for static functions, and do not require a module name prefix either. Team: please opine :+1: or :-1: emoji on the PR description. 

Rationale: Using the `mbedtls_` prefix on static functions makes it easier to export the function later, which is necessary to write unit tests. Not using a prefix is convenient for small auxiliary functions that are very specific to one module.

Current enforcement: low. As of `mbedtls-3.2.1`, 6% of static function names start with `mbedtls_` and 22% aren't prefix with the module name or something close (many of those in `*_wrap.c` prefixed with the algorithm name). Methodology:
```
perl -l -ne '$f = $ARGV; $f =~ s!.*/!!; $f =~ s!\.c$!!; $b = $f; $b =~ s/_.*|parse$|write$//; $b =~ s/^bignum$/mpi/; if (/^static .*\b([A-Za-z0-9]+)\w+\(/) { if ($1 eq "mbedtls") {++$q} elsif ($1 eq $f || $1 eq $b) {++$m} else {++$o} } END { $t = 1. * ($q + $m + $o); $q/=$t; $m/=$t; $o/=$t; printf "Fully qualified: %.02f\nModule prefix: %.02f\nOther: %.02f\n", $q, $m, $o; }' library/*.c
perl -l -ne '$f = $ARGV; $f =~ s!.*/!!; $f =~ s!\.c$!!; $b = $f; $b =~ s/_.*|parse$|write$//; if (/^static .*\b([A-Za-z0-9]+)\w+\(/) { if ($1 ne "mbedtls" && $1 ne $f && $1 ne $b) { print "$f $1" } }' library/*.c
```

Also in this pull request, closely related and non-controversial:

* Clarify that any exported function counts as public. No rule change, just clarification.
* Note that headers in `library/` used by alternative implementations must not define non-namespaced macro names. Change overdue since Mbed TLS 3.0: before that, alternative implementations did not use headers in `library/`.

### DIFF
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -120,7 +120,7 @@ Note that all externally linked functions must have a name starting with `mbedtl
 
 Exception: code implementing the PSA crypto API uses the `PSA_` and `psa_` prefixes. This includes official APIs as well as draft APIs that are on the PSA standards track. API extensions which are meant to remain specific to Mbed TLS, and internal functions, should use the `MBEDTLS_/mbedtls_` prefix, however there are many existing cases of using the `PSA_/psa_` prefix.
 
-Exception: macros may have lowercase names if they behave exactly like normal variables and functions. In particular, function-like macros may have lowercase names only if they expand to an expression that evaluates each argument exactly once.
+Exception: macros may have lowercase names if they behave exactly like normal variables and functions. In particular, function-like macros may have lowercase names only if they expand to an expression that evaluates each argument exactly once. Even in those cases, uppercase names are strongly preferred and new code should not use lowercase macro names for anything other than function replacement (like `#define mbedtls_printf printf`).
 
 ### Local names
 
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -122,9 +122,9 @@ Exception: code implementing the PSA crypto API uses the `PSA_` and `psa_` prefi
 
 ### Local names
 
-Static functions can use the same naming scheme as non-static functions (`mbedtls_MODULE_frobnicate()` or `psa_frobnicate()`). This is convenient if they are made non-static later, for example for testing. They can also have shorter names if it's convenient. Keeping the module name (`MODULE_frobnicate()`) as a prefix makes it makes it slightly easier to follow call stacks across modules, but it is not compulsory.
+Static functions can use the same naming scheme as non-static functions (`mbedtls_MODULE_frobnicate()` or `psa_frobnicate()`). This is convenient if they are made non-static later, for example for testing. They can also have shorter names if it's convenient. Keeping the module name (`MODULE_frobnicate()`) as a prefix makes it slightly easier to follow call stacks across modules, but it is not compulsory.
 
-Macros defined only in the `library` directory should follow the same naming scheme as non-static functions (`MBEDTLS_MODULE_FOO` or `PSA_FOO`).The can use shorter names (omitting `MBEDTLS_`, the module name, or both) if it's convenient. However, avoid names without underscores as some embedded platforms define short macro names in their system headers. Exception: macros defined in headers used by alternative implementations or PSA drivers (including headers that they include) must start with with `MBEDTLS_` or `PSA_`.
+Macros defined only in the `library` directory should follow the same naming scheme as non-static functions (`MBEDTLS_MODULE_FOO` or `PSA_FOO`). They can use shorter names (omitting `MBEDTLS_`, the module name, or both) if it's convenient. However, avoid names without underscores as some embedded platforms define short macro names in their system headers. Exception: macros defined in headers used by alternative implementations or PSA drivers (including headers that they include) must start with `MBEDTLS_` or `PSA_`.
 
 All macros must have uppercase names unless they are function-like (expanding to an expression that evaluates each argument exactly once). Using all-uppercase names for preprocessor constants is recommended. Function-like macros, and macros without arguments that expand to function name, usually follow the lowercase naming convention for functions.
 
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -120,13 +120,13 @@ Note that all externally linked functions must have a name starting with `mbedtl
 
 Exception: code implementing the PSA crypto API uses the `PSA_` and `psa_` prefixes. This includes official APIs as well as draft APIs that are on the PSA standards track. API extensions which are meant to remain specific to Mbed TLS, and internal functions, should use the `MBEDTLS_/mbedtls_` prefix, however there are many existing cases of using the `PSA_/psa_` prefix.
 
+Exception: macros may have lowercase names if they behave exactly like normal variables and functions. In particular, function-like macros may have lowercase names only if they expand to an expression that evaluates each argument exactly once.
+
 ### Local names
 
 Static functions can use the same naming scheme as non-static functions (`mbedtls_MODULE_frobnicate()` or `psa_frobnicate()`). This is convenient if they are made non-static later, for example for testing. They can also have shorter names if it's convenient. Keeping the module name (`MODULE_frobnicate()`) as a prefix makes it slightly easier to follow call stacks across modules, but it is not compulsory.
 
 Macros defined only in the `library` directory should follow the same naming scheme as non-static functions (`MBEDTLS_MODULE_FOO` or `PSA_FOO`). They can use shorter names (omitting `MBEDTLS_`, the module name, or both) if it's convenient. However, avoid names without underscores as some embedded platforms define short macro names in their system headers. Exception: macros defined in headers used by alternative implementations or PSA drivers (including headers that they include) must start with `MBEDTLS_` or `PSA_`.
-
-All macros must have uppercase names unless they are function-like (expanding to an expression that evaluates each argument exactly once). Using all-uppercase names for preprocessor constants is recommended. Function-like macros, and macros without arguments that expand to function name, usually follow the lowercase naming convention for functions.
 
 Function parameters and local variables need no name spacing. They should use descriptive names unless they're very short-lived or are used for simple looping or are "standard" names (such as `p` for a pointer to the current position in a buffer).
 
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -116,6 +116,8 @@ All public names (functions, variables, types, `enum` constants, macros) must st
     mbedtls_aes_setkey_decrypt()
 ```
 
+Note that all externally linked functions must have a name starting with `mbedtls_` to avoid link-time conflicts, even if they are not declared in a public header. This also applies to global variables (which should be used very sparingly).
+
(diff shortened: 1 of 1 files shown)


tom-cosgrove-arm (dismissed):

kb/development/mbedtls-coding-standards.md
```diff
@@ -116,11 +116,15 @@ All public names (functions, variables, types, `enum` constants, macros) must st
...
 ```
 
+Note that all externally linked functions must have a name starting with `mbedtls_` to avoid link-time conflicts, even if they are not declared in a public header. This also applies to global variables (which should be used very sparingly).
+
 Exception: code implementing the PSA crypto API uses the `PSA_` and `psa_` prefixes. This includes official APIs as well as draft APIs that are on the PSA standards track. API extensions which are meant to remain specific to Mbed TLS, and internal functions, should use the `MBEDTLS_/mbedtls_` prefix, however there are many existing cases of using the `PSA_/psa_` prefix.
 
 ### Local names
 
-Static functions and macros that are not in public headers follow the same convention except the initial `MBEDTLS_` or `mbedtls_` prefix: they start directly with the function name.
+Static functions do not need to have the `mbedtls_` prefix. They usually start with the module name (which makes it slightly easier to follow call stacks across modules), but this is not compulsary.
```
davidhorstmann-arm, line 125: Nit: "compulsary" -> "compulsory"

davidhorstmann-arm (approved):
LGTM except one typo - I'll give approval in case it's easier to merge as-is and follow up.

davidhorstmann-arm (dismissed):
LGTM - thanks!

tom-cosgrove-arm (approved):
LGTM, but kicking myself I missed the typo

kb/development/mbedtls-coding-standards.md
```diff
@@ -116,11 +116,15 @@ All public names (functions, variables, types, `enum` constants, macros) must st
...
 ```
 
+Note that all externally linked functions must have a name starting with `mbedtls_` to avoid link-time conflicts, even if they are not declared in a public header. This also applies to global variables (which should be used very sparingly).
+
 Exception: code implementing the PSA crypto API uses the `PSA_` and `psa_` prefixes. This includes official APIs as well as draft APIs that are on the PSA standards track. API extensions which are meant to remain specific to Mbed TLS, and internal functions, should use the `MBEDTLS_/mbedtls_` prefix, however there are many existing cases of using the `PSA_/psa_` prefix.
 
 ### Local names
 
-Static functions and macros that are not in public headers follow the same convention except the initial `MBEDTLS_` or `mbedtls_` prefix: they start directly with the function name.
+Static functions do not need to have the `mbedtls_` prefix. They usually start with the module name (which makes it slightly easier to follow call stacks across modules), but this is not compulsory.
```
yanesca, line 125: I think the module name should be compulsory if the static function in question decides to use the `mbedtls_` prefix.

kb/development/mbedtls-coding-standards.md
```diff
@@ -116,11 +116,15 @@ All public names (functions, variables, types, `enum` constants, macros) must st
...
+
+Macros defined only in the `library` directory should have the `MBEDTLS_` prefix. It's ok to use shorter names if it's convenient, but avoid names without underscores as some embedded platforms define short macro names in their system headers. Exception: macros defined in headers used by alternative implementations or PSA drivers (including headers that they include) must start with with `MBEDTLS_` or `PSA_`.
```
yanesca, line 127: I am not sure I understand. Instead of
> Macros defined only in the `library` directory should have the `MBEDTLS_` prefix.
Did you mean
> Macros defined only in the `library` directory don't need to have the `MBEDTLS_` prefix.

yanesca (changes requested):
I have a couple of questions/suggestions.

kb/development/mbedtls-coding-standards.md
gilles-peskine-arm: Yes, that's what I meant. I'll reword.

kb/development/mbedtls-coding-standards.md
gilles-peskine-arm: I did mean to recommend sticking to the `MBEDTLS_` prefix. We've had problems before with our short names clashing with platform headers that also defined short names (not C compliant, but we had to cope).

yanesca: Ah, I see. Then by

> It's ok to use shorter names if it's convenient

you mean

> It's ok to omit the `MBEDTLS_` prefix if it is convenient

Right?

gilles-peskine-arm: Yes. And that it's also ok to omit the module name (and we do have a few macros already that don't have a module name because they make sense across modules). But keep it sensible and don't define something that's likely to clash like `ASSERT`.

yanesca: Sounds good, could you please clarify this as well when you are doing the update?

tom-cosgrove-arm (approved):
LGTM

kb/development/mbedtls-coding-standards.md
```diff
@@ -116,11 +116,17 @@ All public names (functions, variables, types, `enum` constants, macros) must st
...
+
 Exception: code implementing the PSA crypto API uses the `PSA_` and `psa_` prefixes. This includes official APIs as well as draft APIs that are on the PSA standards track. API extensions which are meant to remain specific to Mbed TLS, and internal functions, should use the `MBEDTLS_/mbedtls_` prefix, however there are many existing cases of using the `PSA_/psa_` prefix.
 
 ### Local names
 
-Static functions and macros that are not in public headers follow the same convention except the initial `MBEDTLS_` or `mbedtls_` prefix: they start directly with the function name.
+Static functions can use the same naming scheme as non-static functions (`mbedtls_MODULE_frobnicate()` or `psa_frobnicate()`). This is convenient if they are made non-static later, for example for testing. They can also have shorter names if it's convenient. Keeping the module name (`MODULE_frobnicate()`) as a prefix makes it makes it slightly easier to follow call stacks across modules, but it is not compulsory.
```
yanesca, line 125:
```suggestion
Static functions can use the same naming scheme as non-static functions (`mbedtls_MODULE_frobnicate()` or `psa_frobnicate()`). This is convenient if they are made non-static later, for example for testing. They can also have shorter names if it's convenient. Keeping the module name (`MODULE_frobnicate()`) as a prefix makes it slightly easier to follow call stacks across modules, but it is not compulsory.
```

kb/development/mbedtls-coding-standards.md
```diff
@@ -116,11 +116,17 @@ All public names (functions, variables, types, `enum` constants, macros) must st
...
+
+Macros defined only in the `library` directory should follow the same naming scheme as non-static functions (`MBEDTLS_MODULE_FOO` or `PSA_FOO`).The can use shorter names (omitting `MBEDTLS_`, the module name, or both) if it's convenient. However, avoid names without underscores as some embedded platforms define short macro names in their system headers. Exception: macros defined in headers used by alternative implementations or PSA drivers (including headers that they include) must start with with `MBEDTLS_` or `PSA_`.
```
yanesca, line 127:
```suggestion
Macros defined only in the `library` directory should follow the same naming scheme as non-static functions (`MBEDTLS_MODULE_FOO` or `PSA_FOO`). They can use shorter names (omitting `MBEDTLS_`, the module name, or both) if it's convenient. However, avoid names without underscores as some embedded platforms define short macro names in their system headers. Exception: macros defined in headers used by alternative implementations or PSA drivers (including headers that they include) must start with with `MBEDTLS_` or `PSA_`.
```

yanesca (dismissed):
Found two typos, but other than that it looks good to me.

kb/development/mbedtls-coding-standards.md
```diff
@@ -116,11 +116,17 @@ All public names (functions, variables, types, `enum` constants, macros) must st
...
+
+All macros must have uppercase names unless they are function-like (expanding to an expression that evaluates each argument exactly once). Using all-uppercase names for preprocessor constants is recommended. Function-like macros, and macros without arguments that expand to function name, usually follow the lowercase naming convention for functions.
```
tom-cosgrove-arm, line 129: > Function-like macros [...] usually follow the lowercase naming convention for functions

The codebase doesn't seem to agree with this.

It's not exact, but:

```
$ find . -name '*.[ch]' | fgrep -v 3rd | xargs perl -n -e 'print if /define [a-z]\w+\(/' | wc
      10      72     586
$ find . -name '*.[ch]' | fgrep -v 3rd | xargs perl -n -e 'print if /define [A-Z]\w+\(/' | wc
     533    3169   32960
```

tom-cosgrove-arm: I've also realised that while the text defines "function-like macros", it defines them to something different to how they are usually understood. e.g. https://www.ibm.com/docs/en/i/7.2?topic=directive-function-like-macros: "Function-like macro definition: An identifier followed by a parameter list in parentheses and the replacement tokens"

Could we have something like "Macros that take parameters must be all-uppercase unless they expand to a direct call to a function, and only evaluate their arguments exactly once"?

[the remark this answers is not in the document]
gilles-peskine-arm: Actually, we did, under “Name spacing”: “Macros and `enum` constants are uppercase with underscores”.

kb/development/mbedtls-coding-standards.md
```diff
@@ -116,11 +116,17 @@ All public names (functions, variables, types, `enum` constants, macros) must st
     mbedtls_aes_setkey_decrypt()
 ```
 
+Note that all externally linked functions must have a name starting with `mbedtls_` to avoid link-time conflicts, even if they are not declared in a public header. This also applies to global variables (which should be used very sparingly).
+
 Exception: code implementing the PSA crypto API uses the `PSA_` and `psa_` prefixes. This includes official APIs as well as draft APIs that are on the PSA standards track. API extensions which are meant to remain specific to Mbed TLS, and internal functions, should use the `MBEDTLS_/mbedtls_` prefix, however there are many existing cases of using the `PSA_/psa_` prefix.
 
+Exception: macros may have lowercase names if they behave exactly like normal variables and functions. In particular, function-like macros may have lowercase names only if they expand to an expression that evaluates each argument exactly once.
```
yanesca, line 123: Remark: I understand that this reflects the reality at this point and therefore this paragraph should be here, so I don't suggest any changes. I just wanted to remark that I am not very happy about this exception and in the future I think we should tighten it (only for macros that evaluate to a function call) or eventually (after we reworked the platform abstraction layer) even remove this exception.

yanesca (approved):
LGTM.

kb/development/mbedtls-coding-standards.md
tom-cosgrove-arm: I would suggest we write what we want to do, i.e. allow lower-case only for macros that evaluate to a function call (e.g. `#define mbedtls_calloc calloc`). There are always exceptions to the coding standards, and so writing something that reflects exactly what we have will means even having to capture inconsistencies. Writing what we would prefer to see seems better, IMO.

gilles-peskine-arm: I've updated the paragraph to say that while it isn't forbidden, it's unwanted.

tom-cosgrove-arm (approved):
LGTM

yanesca (approved):
LGTM

PR #49, opened 2022-09-09
Add new Bignum naming conventions

yanesca:
We agreed on a couple of naming conventions regarding the new Bignum work. This PR is for recording it in the coding standard.

Also added some clarifications to the naming conventions regarding length.

### DIFF
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -130,8 +130,6 @@ By default all lengths and sizes are in bytes (or in number of elements, for arr
 
 `size` should refer to the capacity of a buffer, and `length` to the length of the contents. Most of the time these are interchangeable. A typical exception is when the output is written in a buffer, but the exact length is not known by the caller.
 
-In PSA code `length` and `size` in the name always implies bytes, when referring to number of elements, `count` is used.
-
 ### Modules: `bignum_core`, `bignum_mod` and `bignum_mod_raw`
 
 Generic conventions:
@@ -143,7 +141,7 @@ Generic conventions:
 
 Length parameters:
 
-- For length of `mbedtls_mpi_uint *` buffers we use `limbs` instead of `count`.
+- For length of `mbedtls_mpi_uint *` buffers we use `limbs`.
 - Length parameters are qualified if possible (e.g. `input_length` or `A_limbs`)
 
 ## API conventions
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -128,6 +128,10 @@ Function parameters and local variables need no name spacing. They should use de
 
 By default all lengths and sizes are in bytes (or in number of elements, for arrays). If a name refers to a length or size in bits (as is often the case for key sizes) then the name must explicitly include `bit`, for example `mbedtls_pk_get_bitlen()` returns the size of the key in bits, while `mbedtls_pk_get_len()` returns the size in bytes. In addition, the documentation should always mention explicitly if key sizes are in bits or in bytes.
 
+Size refers to the capacity of the buffer, while length to the length of the contents. Most of the time these are interchangeable. A typical exception is when the output is written in a buffer, but the exact length is not known by the caller.
+
+In PSA code `length` and `size` in the name always implies bytes, when referring to number of elements, `count` is used.
+
 ## API conventions
 
 This section applies fully to classic `mbedtls_xxx()` APIs and mostly to the newer `psa_xxx()` APIs. PSA have their own [conventions described in the PSA Crypto API specification](https://armmbed.github.io/mbed-crypto/html/overview/conventions.html) which take precedence in case of conflicts.
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -132,6 +132,20 @@ Size refers to the capacity of the buffer, while length to the length of the con
 
 In PSA code `length` and `size` in the name always implies bytes, when referring to number of elements, `count` is used.
 
+### Modules: `bignum_core`, `bignum_mod` and `bignum_mod_raw`
+
+Generic conventions:
+
+- `mbedtls_mpi_uint*` operands should be named by capital letters starting at the beginning of the alphabet (`A`, `B`, `C`, ...).
+- `mbedtls_mpi_uint` operands in turn should be named by lower case letters starting at the beginning of the alphabet (`a`, `b`, `c`)
+- For the result `X` or `x` should be used depending on the type.
+- One of the exceptions from this convention is where the naming of function parameters and local variables follows the literature (eg. Handbook of Applied Cryptography)
+
+Length parameters:
+
+- For length of `mbedtls_mpi_uint*` buffers we use `limbs` instead of `count`.
+- Length parameters are qualified if possible (eg. `input_length` or `A_limbs`)
+
 ## API conventions
 
 This section applies fully to classic `mbedtls_xxx()` APIs and mostly to the newer `psa_xxx()` APIs. PSA have their own [conventions described in the PSA Crypto API specification](https://armmbed.github.io/mbed-crypto/html/overview/conventions.html) which take precedence in case of conflicts.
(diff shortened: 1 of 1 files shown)


yanesca:
(Rebased to add sign-off.)

kb/development/mbedtls-coding-standards.md
```diff
@@ -128,6 +128,24 @@ Function parameters and local variables need no name spacing. They should use de
 
 By default all lengths and sizes are in bytes (or in number of elements, for arrays). If a name refers to a length or size in bits (as is often the case for key sizes) then the name must explicitly include `bit`, for example `mbedtls_pk_get_bitlen()` returns the size of the key in bits, while `mbedtls_pk_get_len()` returns the size in bytes. In addition, the documentation should always mention explicitly if key sizes are in bits or in bytes.
 
+Size refers to the capacity of the buffer, while length to the length of the contents. Most of the time these are interchangeable. A typical exception is when the output is written in a buffer, but the exact length is not known by the caller.
+
+In PSA code `length` and `size` in the name always implies bytes, when referring to number of elements, `count` is used.
```
gilles-peskine-arm, line 133: The first comma is technically not grammatical and, more importantly, makes the sentence hard to parse (at first glance, “when” looks like it applies to the first clause).
```suggestion
In PSA code `length` and `size` in the name always implies bytes. When referring to number of elements, `count` is used.
```

gilles-peskine-arm, line 133: That's true in most non-PSA code as well. What existing cases do we have of “size” not being bytes? This paragraph seems mostly redundant with the previous one. PSA has its own [conventions](https://armmbed.github.io/mbed-crypto/html/overview/conventions.html), already linked above, so I'm not sure it's useful to have this PSA-specific paragraph at all.

kb/development/mbedtls-coding-standards.md
```diff
@@ -128,6 +128,24 @@ Function parameters and local variables need no name spacing. They should use de
...
+
+### Modules: `bignum_core`, `bignum_mod` and `bignum_mod_raw`
+
+Generic conventions:
+
+- `mbedtls_mpi_uint*` operands should be named by capital letters starting at the beginning of the alphabet (`A`, `B`, `C`, ...).
```
tom-cosgrove-arm, line 139: Don't we have a space before the `*`? And isn't this specifically for inputs?

```suggestion
- `mbedtls_mpi_uint *` input operands should be named by capital letters starting at the beginning of the alphabet (`A`, `B`, `C`, ...).
```

kb/development/mbedtls-coding-standards.md
gilles-peskine-arm, line 133: Rather than mandate `count` for the number of elements in an array, I think it should be a suggestion when nothing more specific applies. `foos` is a sensible naming pattern for the length of an array of foos, for example it's what bignum does for limb arrays. (A problem with `foos` is that it can also be interpreted as an array of foos, which bignum doesn't do because almost everything is an array of limbs so “this is an array of limbs” is not useful information.)

gilles-peskine-arm (commented):
LGTM on the bignum part, but not so much on the non-bignum part.

kb/development/mbedtls-coding-standards.md
```diff
@@ -128,6 +128,24 @@ Function parameters and local variables need no name spacing. They should use de
...
+- `mbedtls_mpi_uint` operands in turn should be named by lower case letters starting at the beginning of the alphabet (`a`, `b`, `c`)
+- For the result `X` or `x` should be used depending on the type.
+- One of the exceptions from this convention is where the naming of function parameters and local variables follows the literature (eg. Handbook of Applied Cryptography)
+
```
tom-cosgrove-arm, line 143: While this is all well and good to represent `X := A + b * C`, it's less appropriate for things like `montmul` where we have things like `N` for the modulus, `mm` for `mpi_montg_init()` result input and `T` for temporary.

```
- `N` is generally used for the modulus.
- `T` is generally used for a temporary work area given to a function.
- Other specialised parameters (e.g. `mm`) might have their own standardised naming.
- ```
-

File: kb/development/mbedtls-coding-standards.md [hunk 4]
```diff
@@ -128,6 +128,24 @@ Function parameters and local variables need no name spacing. They should use de
 
 By default all lengths and sizes are in bytes (or in number of elements, for arrays). If a name refers to a length or size in bits (as is often the case for key sizes) then the name must explicitly include `bit`, for example `mbedtls_pk_get_bitlen()` returns the size of the key in bits, while `mbedtls_pk_get_len()` returns the size in bytes. In addition, the documentation should always mention explicitly if key sizes are in bits or in bytes.
 
+Size refers to the capacity of the buffer, while length to the length of the contents. Most of the time these are interchangeable. A typical exception is when the output is written in a buffer, but the exact length is not known by the caller.
```
tom-cosgrove-arm, line 131: As with the last sentence of the previous paragraph, I think "should" is a useful word here.
```suggestion
`size` should refer to the capacity of a buffer, and `length` to the length of the contents. Most of the time these are interchangeable. A typical exception is when the output is written in a buffer, but the exact length is not known by the caller.
```

kb/development/mbedtls-coding-standards.md
tom-cosgrove-arm, line 133: This reads better as two sentences:
```suggestion
In PSA code `length` and `size` in the name always implies bytes. When referring to number of elements, `count` is used.
```

kb/development/mbedtls-coding-standards.md
```diff
@@ -128,6 +128,24 @@ Function parameters and local variables need no name spacing. They should use de
...
+In PSA code `length` and `size` in the name always implies bytes, when referring to number of elements, `count` is used.
+
+### Modules: `bignum_core`, `bignum_mod` and `bignum_mod_raw`
+
+Generic conventions:
+
+- `mbedtls_mpi_uint*` operands should be named by capital letters starting at the beginning of the alphabet (`A`, `B`, `C`, ...).
+- `mbedtls_mpi_uint` operands in turn should be named by lower case letters starting at the beginning of the alphabet (`a`, `b`, `c`)
+- For the result `X` or `x` should be used depending on the type.
+- One of the exceptions from this convention is where the naming of function parameters and local variables follows the literature (eg. Handbook of Applied Cryptography)
```
tom-cosgrove-arm, line 142: I personally prefer `e.g.`, and it seems that [english.SE might, too](https://english.stackexchange.com/questions/241104/is-e-g-or-e-g-correct)
```suggestion
- One of the exceptions from this convention is where the naming of function parameters and local variables follows the literature (e.g. Handbook of Applied Cryptography)
```

kb/development/mbedtls-coding-standards.md
```diff
@@ -128,6 +128,24 @@ Function parameters and local variables need no name spacing. They should use de
...
+
+Length parameters:
+
+- For length of `mbedtls_mpi_uint*` buffers we use `limbs` instead of `count`.
```
tom-cosgrove-arm, line 146: Again, isn't the standard to have a space before `*`?
```suggestion
- For length of `mbedtls_mpi_uint *` buffers we use `limbs` instead of `count`.
```

kb/development/mbedtls-coding-standards.md
```diff
@@ -128,6 +128,24 @@ Function parameters and local variables need no name spacing. They should use de
...
+- Length parameters are qualified if possible (eg. `input_length` or `A_limbs`)
```
tom-cosgrove-arm, line 147: e.g. again
```suggestion
- Length parameters are qualified if possible (e.g. `input_length` or `A_limbs`)
```

kb/development/mbedtls-coding-standards.md
yanesca: I have removed this paragraph.

kb/development/mbedtls-coding-standards.md
yanesca: That is why it is should and not must. Anyway, I added `N` and `T` to the guidelines.

(I am not particularly happy about the naming of the init result and I don't think it is a generic enough to be captured here.)

tom-cosgrove-arm (approved):
LGTM

gilles-peskine-arm (approved):

PR #52, opened 2022-09-27
Add how to documentation for bignum test data generation

wernerlewis:
Adds a knowledge base page with guidance on how to use the test generation framework.

This includes some information on how the classes are structured, and a step-by-step examples for adding a test script and adding test case generation for a function.


kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,329 @@
+# How to systematically generate test data
+
+When adding new test functions or improving coverage of existing tests in Mbed TLS, a large number of test cases may need to be added. This can be time-consuming and error-prone when adding manually. Mbed TLS includes a Python framework which can be used to systematically generate test case data from input values. The framework is flexible, to support a variety of test functions, and uses inheritance to reduce code repetition in similar test functions.
```
tom-cosgrove-arm, line 3: I personally find long lines hard to read, even when the editor wraps them. Typically I find 90-100 characters a good maximum.

In addition, I also tend to start a new line when there's a new sentence - when this isn't done, a later PR changing the text will often cause reformatting of all the following lines, which makes diffs harder to read than they need to be.

wernerlewis: This is definitely an oversight, I'll add a commit to split the lines up.

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Creating a test script
+
+This is only required if an existing script is not being extended. An example is included for each step, showing how to create a test generation script for `test_suite_mpi.function`, a basic equivalent of [generate_bignum_tests.py](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/generate_bignum_tests.py).
+
+### Initial Python script
+
+To create the script, `test_generation` must be imported from `mbedtls_dev` and `test_generation.main()` called when the script is ran. The `scripts/` directory must be added to the system path before importing from `mbedtls_dev`; if the new script is in `tests/scripts/` this can be done by importing `scripts_path`.
+
+For the example, create `generate_bignum_ex_tests.py` in `tests/scripts/`, containing:
```
tom-cosgrove-arm, line 119: Could we have a blank line here please?

```suggestion
For the example, create `generate_bignum_ex_tests.py` in `tests/scripts/`, containing:

```

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+This class must implement all attributes and methods of `BaseTarget` and will generate
+test cases.
+This is a [test function class](#test-function-classes).
+
+As an example, test case generation is added for the test function `mpi_add_mpi()`.
+The added class will be derived from `BignumTarget`, and can be added either to
+[tests/scripts/generate_bignum_tests.py](https://github.com/Mbed-TLS/mbedtls/blob/development/tests/scripts/generate_bignum_tests.py)
+or the example script described in the previous section.
+
+### Creating the class
```
gilles-peskine-arm, line 192: For the sake of people who aren't reading the document linearly, please clarify “the class” in the title. Even just “Creating the function class” would be a lot clearer.

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+    test_name = "MPI add"
+```
+
+### Class initialization
+
+Each instance of the class will represent a different test case, each with unique inputs.
+To initialize an instance of the class, these inputs must be passed to the `__init__()`
+constructor method.
+
+For `mpi_add_mpi()`, two input values are required, `val_a` and `val_b`.
```
gilles-peskine-arm, line 220: Almost everything in this paragraph is specific to generating bignum tests, and irrelevant otherwise. I think it would be more helpful to have a non-bignum or simplified example that doesn't go into these considerations about leading zeros, and a separate more advanced example that does illustrate how to deal with bignums. The more advanced example could be just a statement to go and look at a particular class in `generate_bignum_tests.py` though.

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+        # int(val, 16) converts from a hex string to int
+        # empty strings should be treated as zero in calculations
+        self.int_a = int(val_a, 16) if val_a else 0
+        self.int_b = int(val_b, 16) if val_b else 0
+```
+
+### Implementing abstract BaseTarget methods
+
+Abstract methods declared in `BaseTarget` must also be implemented in test function
+classes.
```
gilles-peskine-arm, line 244: I'm not sure what that sentence means — I guess this?
```suggestion
The abstract methods declared in `BaseTarget` must have an implementation in test function
classes.
```
which could be simplified to
```suggestion
Test function classes must provide an implementation of all the abstract methods declared in `BaseTarget`.
```

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+For the example, the following imports will be required:
+
+```python
+import itertools
+from typing import Iterator, List
+# After import scripts_path
+from mbedtls_dev import test_case
+```
+
+`arguments()` returns the list of arguments required by the test function.
```
gilles-peskine-arm, line 255:
```suggestion
The method `arguments()` returns the list of arguments passed to the test function.
```
(“required by” makes me think it returns some kind of specification that describes what the function requires, but that's not it.)

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+Any C-string arguments must be quoted.
```
gilles-peskine-arm, line 256: I'm not sure what “C-string arguments” refers to here. Is this about test function arguments that are passed to test functions as `char *`? If so this statement is incomplete: `data_t` arguments also need to be quoted. It would be clearer to formulate this differently, something like

> Arguments must have the syntax expected in `.data` files, with double quotes around arguments that are strings or hex data.

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+In the example, a `result()` method is also added to calculate the expected output.
```
gilles-peskine-arm, line 257: Suggested reformulation:
```suggestion
In the example, we also use an auxiliary method `result()` to calculate the expected output.
```
This makes it clearer that _we_ (the author(s)) are choosing to use a method called `result`, it isn't something that the framework requires.

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+            # Initialize a class instance with these values
+            class_instance = cls(a_value, b_value)
+            # Yield the TestCase object from this instance
+            yield class_instance.create_test_case()
+```
+
+### Other BaseTarget attributes and methods
+
+`description()` returns a description of the test case, in the format "`test_name`
+#`count` `case_description`" by default.
```
gilles-peskine-arm, line 302: Don't put `#` as the first character of a line: in many markdown variants, it starts a title even when the previous line is not blank.

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+                "0 (null)" if self.arg_b == "" else hex(self.int_b)
+            )
+        # Call description() in the parent class
+        return super().description()
+```
+
+`dependencies` is a list of build macros, which specifies the required build config for
+the test case to run.
+This can be set as a class attribute if always required for the test function, or set
+conditionally in the constructor if dependent on the test case.
```
gilles-peskine-arm, line 325: This warrants an example: how do you set a dependency for one specific test case?

(Setting a dependency for a whole function is rarely useful, since that can go in the `.function` file. It's only useful when there are multiple function classes with the same function name.)

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+
+### Variants
+
+For variants of a test function, inheritance can be used to reduce repetition of code.
+For example, to implement the test function `mpi_add_abs()`, `BignumAddAbs` can inherit
+from the `BignumAdd` class.
+The necessary changes for implementing `BignumAddAbs` are then reduced:
+ - `count` should be reset to 0 for the class.
+ - `test_function` should be set to `mpi_add_mpi_abs`.
+ - `test_name` should be set.
```
gilles-peskine-arm, line 433:
```suggestion
 - `test_name` should be set to a different name.
```

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+ - `int` forms of the inputs (`int_a` and `int_b`) should be made absolute.
```
gilles-peskine-arm, line 434:
```suggestion
 - We take the same list of inputs as the parent class, but calculate the expected result differently.
```

kb/how-to/add-test-generation.md
```diff
@@ -0,0 +1,478 @@
...
+    """Tests for absolute variant of bignum addition."""
+    count = 0
+    test_function = "mpi_add_abs"
+    test_name = "MPI add (abs)"
+
+# Either override __init__()
+    def __init__(self, val_a, val_b) -> None:
+        super().__init__(val_a, val_b)
+        self.int_a = abs(self.int_a)
+        self.int_b = abs(self.int_b)
```
gilles-peskine-arm, line 449: I guess this works due to the way `BignumAdd` is structured, because `__init__` stores the representation of the A and B arguments in fields that aren't `int_a` and `int_b` but does not itself calculate the representation of the expected-result argument. But if the reader doesn't know that, they'll be completely lost. Besides, it's confusing to show the same piece of code with three parts where you're only supposed to take parts 1+2 or 1+3. I suggest not showing this approach at all, or in a different example.

gilles-peskine-arm (changes requested):
Thanks for writing this guide! I found a few parts unclear though.

PR #54, opened 2022-10-07
Replace "precompiler" with "preprocessor" in docs

davidhorstmann-arm:
Substitute the more commonly understood term "preprocessor" where currently the equivalent but rarer term "precompiler" is used in the documentation.


mpg (approved):
LGTM

PR #58, opened 2022-10-21
Coding style: formatting of lists

gilles-peskine-arm:
* Function calls: group arguments sensibly
* Initializers and enum: one item per line, or sensible groups of numerical values
* Include trailing commas

This would also apply to scripts, but [PEP8](https://peps.python.org/pep-0008/#when-to-use-trailing-commas) covers trailing commas, so I don't think we need anything explicit in the script coding style.

### DIFF
--- a/kb/development/mbedtls-coding-standards.md
+++ b/kb/development/mbedtls-coding-standards.md
@@ -99,6 +99,13 @@ Similarly, sizeof expressions always use parentheses even when it is not necessa
 
 ### Formatting of lists
 
+When a function or macro call doesn't fit on a single line, put one argument per line or a sensible grouping per line. For example, in the snippet below, `input_buffer` and `input_size` are on a line of their own even though the call could fit on two lines instead of three if they were separated:
+```c
+    function_with_a_very_long_name( parameter1, parameter2,
+                                    input_buffer, input_size,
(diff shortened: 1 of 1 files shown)


kb/development/mbedtls-coding-standards.md
```diff
@@ -97,6 +97,30 @@ Similarly, sizeof expressions always use parentheses even when it is not necessa
...
+When a function or macro call doesn't fit on a single line, put one argument per line or a sensible grouping per line. For example, in the snippet below, `input_buffer` and `input_size` are on a line of their own even though the call could fit on two lines instead of three if they were separated:
+```c
+    function_with_a_very_long_name( parameter1, parameter2,
+                                    input_buffer, input_size,
+                                    output_buffer, output_size );
+```
+
+Lists of items other than function arguments should generally have one item per line. Exceptions:
+
+* It's ok to write simple structure initializers on a single line.
```
gilles-peskine-arm, line 111: I'm not saying anything about designated initializers because we currently don't use them yet, because we support VS2013 which IIRC doesn't support them.

tom-cosgrove-arm (dismissed):
LGTM

tom-cosgrove-arm (approved):

PR #64, opened 2022-11-09
Python venv setup guide

gilles-peskine-arm:
This is intended to be comprehensible by someone with no prior experience of Python package management.

### DIFF
--- a/kb/development/python-environment.md
+++ b/kb/development/python-environment.md
@@ -1,0 +1,50 @@
+# Setting up a Python environment for Mbed TLS
+
+## Python and package versions
+
+Each branch of Mbed TLS has different minimum requirements on Python. See `README.md` in the branch. Any Python version that's more recent than the minimum should work, however as of November 2022 this is not well tested by the CI so unintended incompatibilities can happen.
+
+Some of the scripts used by Mbed TLS users and developers require third-party packages. Those have version constraints as well. The constraints are documented via `*.requirements.txt` files in the [`scripts` directory](https://github.com/Mbed-TLS/mbedtls/tree/development/scripts).
+
+We run [mypy](http://mypy-lang.org/) and [pylint](https://pylint.pycqa.org/en/latest/) to check Python scripts. It is common that newer versions of Pylint reject our code, because some checks have become stricter or new checks have been added.
+
+## Set up a Python package environment
+
+### Quick setup
+
+Initial setup:
+
+```sh
+python3 -m venv ~/venv-mbedtls-development
+~/venv-mbedtls-development/bin/python scripts/min_requirements.py
+```
+
+Day-to-day work: start a shell and issue this command:
+```sh
+. ~/venv-mbedtls-development/bin/activate
+```
+
+### Detailed explanations
+
+From any installation of Python, you can set up a [virtual environemnt](https://realpython.com/python-virtual-environments-a-primer/) (“venv” for short) with a specific set of packages. When working in this environment, none of the third-party packages installed globally are visible, so you can have specific versions in this environment without contaminating, or getting contaminated by, your normal system.
+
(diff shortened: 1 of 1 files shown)


kb/development/python-environment.md
```diff
@@ -0,0 +1,50 @@
...
+Initial setup:
+
+```sh
+python3 -m venv ~/venv-mbedtls-development
+~/venv-mbedtls-development/bin/python scripts/min_requirements.py
+```
+
+Day-to-day work: start a shell and issue this command:
+```sh
+. ~/venv-mbedtls-development/bin/activate
```
gilles-peskine-arm, line 24: Should you unset `PYTHONPATH` as well? This can cause disruption sometimes.

kb/development/python-environment.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Set up a Python package environment
+
+### Quick setup
+
+Initial setup:
+
+```sh
+python3 -m venv ~/venv-mbedtls-development
+~/venv-mbedtls-development/bin/python scripts/min_requirements.py
```
yanesca, line 19: Is this different from
```sh
python3 -m venv ~/venv-mbedtls-development
. ~/venv-mbedtls-development/bin/activate
python scripts/min_requirements.py
```

gilles-peskine-arm: It's the same. Running `activate` in a shell can't really be undone, so I don't want to show instructions to run it without starting a separate shell.

yanesca: Shouldn't `deactivate` undo it?

kb/development/python-environment.md
yanesca: I don't think that should be in the basic instructions, but might worth mentioning in a remark.

yanesca (approved):
Left some questions/comments, but looks good to me already as it is.

kb/development/python-environment.md
minosgalanakis: With regards to PYTHONPATH I believe nowadays venv is unsetting `PYTHONHOME`. 

But there is nothing stopping users from adding an extra unset for it in ~/venv-mbedtls-development/bin/activate

It falls outside of the scope of this document though. If needed we could have a remark in line of 

> If required `$PYTHONPATH` and `$PYTHONHOME` can be configured or unset by modifying `~/venv-mbedtls-development/bin/activate`

kb/development/python-environment.md
gilles-peskine-arm: Ah, it should. I'm not in the habit of using it, I don't remember if it's because it used to not exist or it used to not work well or I just don't like the risk when starting a new shell is so easy.

kb/development/python-environment.md
```diff
@@ -0,0 +1,64 @@
...
+```sh
+. ~/venv-mbedtls-development/bin/activate
+```
+The same `activate` script works in zsh and ksh as well. There are also `activate.*` scripts for fish, csh and powershell.
+
+Now you can run e.g. `tests/scripts/check-python-files.sh` with the same tool versions as the CI.
+
+### Installing different Python versions
+
+Each virtual environment can have a different Python version, and running in a virtual environment makes that version of Python the one on the command search path. But all the Python versions need to be installed at the system level.
```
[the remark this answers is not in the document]
gilles-peskine-arm: The previous section already discusses setting up separate virtual environments with different python versions, so I didn't think it was worth going in detail here. But it is a little hard to follow in the previous section as well, so I'll massage the text a bit.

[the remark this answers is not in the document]
gilles-peskine-arm: `venv` is part of the standard library. You don't need a separate package. The Debian `python3.N-venv` package just has a wrapper script that runs `venv.main()`.

minosgalanakis (approved):
LGTM

yanesca (approved):
LGTM

PR #66, opened 2022-11-16
Mention retiring legacy crypto API as a plan

mpg:
It's been on our minds for a while now, and is already mentioned publicly in a few places (like `docs/architecture/psa-migration/strategy.md`) not discoverable or clearly explained, so let's fix that.


gilles-peskine-arm (approved):
LGTM

PR #72, opened 2023-01-10
Fix branch rewrite script issue with deleted files

davidhorstmann-arm:
Fix a problem with the branch rewrite script that prevented it working when the branch deletes files. Whenever there are deleted files, manually `git rm` them to resolve conflicts with `git cherry-pick`.  Add a testcase for this eventuality.

Additionally, add 1 unrelated real-world testcase taken from Mbed-TLS/mbedtls#6882.

### DIFF
--- a/tools/bin/mbedtls-rewrite-branch-style
+++ b/tools/bin/mbedtls-rewrite-branch-style
@@ -411,6 +411,20 @@ class CodeStyleBranchRewriter(MbedTLSBranchRewriter):
                      if line[0] != 'D'] # skip deleted files
         return filenames
 
+    def deleted_files_in_commit(self, revision: str) -> List[str]:
+        """The list of files changed or added in the given revision."""
+        # The git command prints a list of the form:
+        #   * 'A\t{filename}' for an added file
+        #   * 'M\t{filename}' for a modified file
+        #   * 'D\t{filename}' for a deleted file
+        #   * 'R{confidence}\t{old_name}\t{new_name}' for a renamed file
+        lines = self.get_git_lines(['show', '--format=', '--name-status',
+                                    revision])
+        deleted = [line.rsplit('\t', 1)[-1]
+                     for line in lines
+                     if line[0] == 'D']
+        return deleted
+
     @staticmethod
     def is_c_file(path: pathlib.PurePath) -> bool:
         """Whether path is a C source file, based on its name and location."""
@@ -466,6 +480,11 @@ class CodeStyleBranchRewriter(MbedTLSBranchRewriter):
     def restyle_commit_onto_current(self, new_commit: str) -> None:
         """Apply a restyled content of new_commit onto HEAD."""
         self.info('Applying {} onto {}', new_commit, self.head_sha())
+        # If the commit deletes files that are restyled there will be
+        # a merge conflict. Manually get a list of the deleted files that
+        # are eligible for restyling.
+        deleted_files = self.deleted_files_in_commit(new_commit)
+        deleted_styled_files = self.filter_styled_files(deleted_files)
         # We want the new commit's metadata (author, commit message),
         # the new commit's content (but restyled), and the old commit's
         # history. This is a cherry-pick operation, with some custom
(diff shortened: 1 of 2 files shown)


tools/bin/mbedtls-rewrite-branch-style
```diff
@@ -477,7 +496,23 @@ class CodeStyleBranchRewriter(MbedTLSBranchRewriter):
         # cherry-picked commit, but might mess up modified files. Then
         # we override the content of modified files, and finally we restyle
         # files that need restyling.
-        self.run_git(['cherry-pick', '-Xtheirs', '--allow-empty', new_commit])
+        if deleted_styled_files:
+            # There are some deleted files, so expect cherry-pick to return
+            # with exit status 1.
+            try:
+                self.run_git(['cherry-pick', '-Xtheirs', '--allow-empty', new_commit])
+            except subprocess.CalledProcessError as c:
```
gilles-peskine-arm, line 504: I don't like catching an error this way, because it could also catch other errors. Can we make `cherry-pick` not error out if there are deleted files?

gilles-peskine-arm: [Passing `--keep-redundant-commits` to `cherry-pick`](https://github.com/Mbed-TLS/mbedtls-docs/pull/71#discussion_r1066118119) is enough to make `test_6882` pass, but not `teest_rename_delete`. Given that the real-world case passes, I'm inclined to consider that good enough if we can't find a better solution quickly.

davidhorstmann-arm: Ah I should perhaps have documented more clearly - `test_6882` is just another testcase I added to increase the total number, it's unrelated to deleted files.

davidhorstmann-arm: To make `cherry-pick` not return an error would require some kind of `--return-success-on-conflicts` command-line switch, which I've looked for and not found. What I have done is rethrown the error if the return value is anything other than 1, which will catch some if not all other errors.

gilles-peskine-arm: Ah, my bad, I should have looked.

Maybe the right thing would be to assemble the commit object manually... That's a bit too low-level for my tastes though.

Alternatively, perhaps try `ours` as the merge strategy, then figure out what to do for added and deleted files (including renames/copies)?

The right solution is actually probably to treat this as an interactive rebase and give the user the opportunity to resolve the conflict manually. That requires the script to carry on with its original state though. I'd actually have liked to make the process `git rebase --exec=/path/to/our/script`, but I couldn't figure out how to arrange what I want this way.

davidhorstmann-arm: Hmm
I suppose thanks to the `reset`-ing later on, the actual state of the tree after the `cherry-pick` does not matter, we're just doing it to retain commit details. In that case I suppose `-Xours` might fix the whole issue (since it cannot fail, as far as I'm aware).

tools/bin/mbedtls-rewrite-branch-style
```diff
@@ -477,7 +496,23 @@ class CodeStyleBranchRewriter(MbedTLSBranchRewriter):
...
+                if c.returncode != 1:
+                    raise c
+            # Resolve conflicts by removing any deleted files then continue.
+            self.run_git(['rm'] + deleted_styled_files)
+            # Continue the cherry-pick with the editor set to the 'true'
+            # command which returns success immediately. In other words,
+            # don't open an editor.
+            self.run_git(['cherry-pick', '--continue'],
+                         env={**os.environ, 'GIT_EDITOR':'true'})
```
gilles-peskine-arm, line 513: I think `GIT_EDITOR=true` won't work on Windows. Unfortunately `git cherry-pick` lacks a `--no-edit` option except in very recent git versions. I think `git commit --no-edit --allow-empty` will work here, except it adds a `# Conflicts:` block to the log message (we'd have to delete `MERGE_MSG` to avoid that).

tools/bin/mbedtls-rewrite-branch-style
gilles-peskine-arm: According to my notes, `-Xtheirs`

> ensures that added and removed files are according to the cherry-picked commit, but might mess up modified files

but I'm not fully confident. In fact, the fact that my code didn't work with deleted files shows that it isn't quite doing what I think it does.

gilles-peskine-arm: I've been experimenting a little with `--strategy=ours` (I think you meant this — `-Xours` can fail, and in fact does fail just like `-Xtheirs` in the “Stop testing Aria” commit). That does avoid conflicts, but I don't see an easy way to handle added/renamed files (and likely mode changes would disappear, too).

davidhorstmann-arm: Ah yes I think that is what I meant. I changed my test copy to that but as you say, it fell over on a rename. It might be fixable by reordering the `git reset`s somehow, but I'm not quite sure how.

Alternatively, we could take the loss of some error reporting as an acceptable tradeoff given the complication of other approaches and stick with what is in this PR.

gilles-peskine-arm: What bothers me is that we're losing error _detection_, not just error _reporting_. As in, we're potentially quietly constructing wrong output, and not just erroring with an incomprehensible message.

I'm looking for a way to say “if there's a deleted file in their branch, just delete it” but no luck so far.

gilles-peskine-arm: Hmmm, maybe delete the file _before_ the cherry-pick? And then squash the cherry-pick commit, but arranging to pick the cherry-pick commit's non-content data rather than the delete commit.

gilles-peskine-arm: > maybe delete the file before the cherry-pick?

Done in https://github.com/Mbed-TLS/mbedtls-docs/pull/74

gilles-peskine-arm:
Superseded by https://github.com/Mbed-TLS/mbedtls-docs/pull/74

PR #73, opened 2023-01-10
Branch rewrite script: minor improvements

gilles-peskine-arm:
Some minor follow-ups to https://github.com/Mbed-TLS/mbedtls-docs/pull/71:

* Fix git detection on Windows.
* More relaxed detection of upstream repository.
* Fix some cases of a commit becoming empty.

PR #74, opened 2023-01-10
Branch rewrite script: handle deleted files

gilles-peskine-arm:
Fix a failure of the branch rewrite script when a C file is deleted: this caused a modify/delete confict during cherry-pick.

I use an alternative approach compared with #71 which avoids creating an error situation where we ignore an error, but aren't sure that the error we're ignoring is the error we were expecting.

This pull request includes the non-controversial changes in https://github.com/Mbed-TLS/mbedtls-docs/pull/73 because I haven't bothered isolating the tests and managing the trivial add/add conflicts that rebasing would cause. The first new commit is "Fix handling of commits that delete a C file".


davidhorstmann-arm (approved):
LGTM - thanks!

PR #88, opened 2023-03-30
Add the new release checker script

paul-elliott-arm:
This is the first working version of the release checker script, that was used to generate the 3.4.0 / 2.28.3 releases.


tools/bin/mbedtls-release-checker.py
```diff
@@ -0,0 +1,474 @@
...
+                    csv_row.append(self.get_cvs_link('https://github.com/{}/pull/{}'.format(self.github_repo,
+                                                                                            link_pr),
+                                                     link_pr))
+
+                csv_writer.writerow(csv_row)
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Attempt to automatically detect missing backports/Changelog entries for an MbedTLS release')
+    parser.add_argument('LTS_range_start', help='Git hash/label of LTS branch range start')
```
minosgalanakis, line 420: I would reword it or make it clear in the help section that this is an immutable ref tag.

Also this could have a default set by `git describe --abbrev=0 origin/development`

tools/bin/mbedtls-release-checker.py
```diff
@@ -0,0 +1,474 @@
...
+    parser.add_argument('LTS_range_end', help='Git hash/label of LTS branch range start')
```
minosgalanakis, line 421: argparser is using ordereddict so the order of inclusion of the arguments here determines how the will be assigned when parsed. 

For example using the syntax described in the docstring `release-checker.py development-local mbedtls.prev.release mbedtls-lts mbedtls.prev.lts.release` and printing the vars(args) I noticed that: 


~~~~~~
/mbedtls-release-checker.py -vv development-local mbedtls-3.4.1 mbedtls-2.28 v2.28.4
{'LTS_output': None,
 'LTS_range_end': 'mbedtls-3.4.1',
 'LTS_range_start': 'development-local',
 'dev_output': None,
 'dev_range_end': 'v2.28.4',
 'dev_range_start': 'mbedtls-2.28',
 'restricted': False,
 'verbose': False,
 'veryverbose': True}
~~~~~~

Ideally I would adjust the order to be like like that `mbedtls.prev.release` - `mbedtls.dev.branch` - `mbedtls.prev.lts.release` - `mbedtls.lts.branch`

Also I would use the standard long and short notation for all arguments.
e.g.

~~~~~~
def get_latest_tag(lts=False):
    """ Return the latest tags for release and lts releases """

    cmd = ["git","describe","--abbrev=0"]
    cmd += ["origin/mbedtls-2.28"] if lts else ["origin/development"]
    out = subprocess.check_output(cmd) # Add try/catch logic
    return out.decode('UTF-8').strip()

 parser.add_argument("-ltr", "--lts-release", default=get_latest_tag(lts=True), help="Tag pointing to the latest LTS release",`
 parser.add_argument("-ltb", "--lts-branch", default="mbedtls-2.28", help="Branch containing all the code to be included in the new release",
~~~~~~

tools/bin/mbedtls-release-checker.py
```diff
@@ -0,0 +1,474 @@
...
+    parser.add_argument('dev_range_start', help='Git hash/label of development branch range start')
+    parser.add_argument('dev_range_end', help='Hash/label of development branch range end')
+    parser.add_argument('-lo', '--LTS_output', help='Ouput file to dump LTS release csv data to')
+    parser.add_argument('-do', '--dev_output', help='Ouput file to dump development release csv data to')
```
minosgalanakis, line 425: I would have predefined defaults for those files. This script takes a significant ammount and API calls to go through a run, that will do nothing in the current default configuration. 

It would make more sense for the user to explicitly turn off output file generation.

minosgalanakis (commented):
First round of reviews with several comments/improvements. I will be following up with a second round for more details on the code flow.

PR #90, opened 2023-04-14
Redirect previous roadmap location to current one

davidhorstmann-arm:
The document `roadmap.md` was linked-to several times then moved as part of restructuring work. In order to retain the working link, add the `sphinx-reredirects` module and tell it to redirect the old `roadmap.md` location to the new one.

PR #91, opened 2023-04-14
Add post-build step to update redirects

bensze01:
This PR allows us to maintain our page redirects in-tree.
The update step only runs when the `latest` version is built ie. after a PR is merged.

See #90 for previous discussion.

RFC: Does the team consider this useful? Or is everyone content with managing redirect separately?


bensze01:
[Test run](https://readthedocs.org/projects/mbed-tls/builds/20118716/) (on a separate `test` branch, and with the branch filter modified, since the API token is not exported in PR jobs for safety reasons.)

mpg:
> RFC: Does the team consider this useful?

Yes! Please let's have as much in-tree as possible!

davidhorstmann-arm:
> the API token is not exported in PR jobs for safety reasons

Would you mind documenting this (maybe in the `redirects.yaml`?) so that people who raise PRs to fix redirects know not to expect the PR build to have the redirect?

PR #92, opened 2023-04-18
Clarify priorities and add priority-very-high

daverodgman:
Add note for `priority-very-high`; clarify wording for `priority-high`.

### DIFF
--- a/reviews/review-priorities.md
+++ b/reviews/review-priorities.md
@@ -16,12 +16,16 @@ Classifications are done through labels, e.g.: priority-high.
 
 Please note that these are the priorities used by the core team. External contributors are welcome to review PRs according to their own priorities, and such review activity will help highlight PRs which should get attention from the core team.
 
+## priority-very-high
+
+Same criteria as priority-high, below, but needed for the Mbed TLS teams' most important commitments.
+
 
 ## priority-high
 
-This classification is used for PRs which are typically one of:
+This classification is used for PRs which have a target date for merging - typically one of:
 
-- Security (or other high-priority) bug fixes
+- Security (or other high-priority) bug fixes needed in the next release
 - Items that the Mbed TLS team have planned for delivery in the current quarter
 
 These PRs should receive review promptly.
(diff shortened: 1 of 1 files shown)


reviews/review-priorities.md
```diff
@@ -16,12 +16,16 @@ Classifications are done through labels, e.g.: priority-high.
...
+Same criteria as priority-high, below, but needed for the Mbed TLS teams' most important commitments.
+
 
 ## priority-high
 
-This classification is used for PRs which are typically one of:
+This classification is used for PRs which have a target date for merging - typically one of:
 
-- Security (or other high-priority) bug fixes
+- Security (or other high-priority) bug fixes needed in the next release
```
yanesca, line 28: Security fixes needed in the next release are mostly `priority-very-high` as they block releases, force releases and often have a hard deadline for public disclosure. Maybe we could mention them as an example in the very high category.

reviews/review-priorities.md
```diff
@@ -16,12 +16,16 @@ Classifications are done through labels, e.g.: priority-high.
 
 Please note that these are the priorities used by the core team. External contributors are welcome to review PRs according to their own priorities, and such review activity will help highlight PRs which should get attention from the core team.
 
+## priority-very-high
+
+Same criteria as priority-high, below, but needed for the Mbed TLS teams' most important commitments.
```
yanesca, line 21: I feel that a forward reference is not very reader friendly. Also, as far as I understand the difference between the two is that `priority-high` is only a goal, whereas `priority-very-high` is a commitment. How about something like this:
> This classification is used for PRs which have a committed deadline for merging.

yanesca (commented):
Left a couple of comments inline.

reviews/review-priorities.md
daverodgman: They are only `priority-very-high` if the release itself is a diamond commitment, which usually it isn't (although maybe that should change).

reviews/review-priorities.md
tom-cosgrove-arm, line 21: Could probably go with the simpler "the Mbed TLS team's" (i.e. possessive of singular team, rather than plural teams)

daverodgman: `priority-very-high` is for diamond commitments (I haven't used that phrase here as it's meaningless outside of Arm / OSS), whereas `priority-high` is for standard commitments. Both have a deadline; the difference is that if a diamond is at risk, we might choose to take resource away from a non-diamond commitment to help.

But from an external point of view, they are pretty similar.

daverodgman: Agree the forward reference isn't ideal, but it's tricky to avoid it without restructuring the whole thing.

tom-cosgrove-arm:
> fix apostrophe's

😱

yanesca (approved):
LGTM

tom-cosgrove-arm (approved):
LGTM

PR #93, opened 2023-04-21
Document the review process for contributors

mpg:
I've written parts of this document dozens of time over the years as github comments on someone's PR, and I'm sure I'm not alone. Let's get this properly documented.


reviews/review-for-contributors.md
```diff
@@ -0,0 +1,45 @@
+# Mbed TLS Review Guidelines for Contributors
+
+## Introduction
+
+This document describes the process of having your pull request (PR) reviewed.
```
gilles-peskine-arm, line 5: Link to [CONTRIBUTING.md](https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md) here?

reviews/review-for-contributors.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Creating backports
+
+Some changes need to be backported to the Long-Term Support (LTS) branches: security fixes, bug fixes, and some testing improvements. In that case, as a general rule we wait until the backport is ready and approved before merging the main PR.
+
+A backport is a separate PR that, instead of targeting `development` as usual, targets a LTS branch. That secondary PR should have the same title as the main one with "[Backport x.yy] " prepended, where x.yy is the name of LTS branch (for example 2.28). The description should start with "This is the x.yy backport of #zzzz", where zzzz is the number of the original PR, and then mention if the backport is trivial, or if there were conflicts that had to be resolved, or important differences with the main branch (for example: "only contextual conflicts", or "had to add/remove XXX because feature YYY is present only on one side", or "skipped commits U and V because they were irrelevant").
```
gilles-peskine-arm, line 41: > with "[Backport x.yy] " prepended

Minor: we use "Backport x.yy:" as the prefix about three times as often (stats from 2.28).

jethrogb:
This should probably be linked from https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md

mpg:
> This should probably be linked from https://github.com/Mbed-TLS/mbedtls/blob/development/CONTRIBUTING.md

Absolutely. https://github.com/Mbed-TLS/mbedtls/pull/7468

reviews/review-for-contributors.md
```diff
@@ -0,0 +1,45 @@
...
+* Once you have addressed one point of feedback, please leave the GitHub discussion unresolved: the person who raise the point should mark it as resolved if they are satisfied with the resolution. This allows reviewers to easily track if all of their feedback as been addressed.
+  * If you want a way to mark which points you think you addressed, we tend to do that by reacting with the `:rocket:` emoji to the original comment. (We also use the thumbs up emoji as a way of saying we agree with the comment and intend to address it.) Note that this is purely optional.
+* Please avoid force-pushing once review has started, except in the circumstances listed below; this tends to disrupt incremental review and is not supported well enough by GitHub's reviewing tools. In particular, never force-push several times in a row; if you did, GitHub would to merge the events in the PR's timeline, showing only details of the last force-push, making it impossible to find again which commit was last reviewed. Exceptions:
```
gilles-peskine-arm, line 30: > GitHub would to merge the events in the PR's timeline, showing only details of the last force-push

That's not completely true, and there have been changes in the amount of information GitHub changes. As far as I remember, it's always been the case that the web interface (but not the API :frowning_face:) makes the history's history accessible if you dig enough, provided that some non-push even happens between two successive force-pushes (where “between” might require a few minutes on either side due to internal caching).

> never force-push several times in a row

I don't think we should forbid it, and in practice we do it often when the circumstances are right. For example:

1. Oh, there's a CI failure for something I can't test locally. Ok, I think this'll fix it. &lt;hackity&gt; &lt;push&gt; &lt;wait&gt;
2. Oh no, the fix didn't quite work right. &lt;hackity&gt; &lt;force-push&gt; &lt;wait&gt;
3. Repeat until CI passes.

I think force-pushing is generally ok _as long as nobody has responded (commented, or starting a review) to the commit you'll be overriding_, and as long as you're not squashing away useful insights about the history of the code.

(Case in point about “as long as nobody has responded”: GitHub prevented me from posting this comment when I was done writing it because you just force-pushed to rewrite the commit message…)

reviews/index.md
```diff
@@ -1,5 +1,6 @@
 # Reviews
 ```{toctree}
 review_guidelines.md
+review-for-contributors.md
```
davidhorstmann-arm, line 4: I think we should put the contributors doc first as we have more contributors than reviewers.
```suggestion
review-for-contributors.md
review_guidelines.md
```

reviews/review-for-contributors.md
```diff
@@ -0,0 +1,45 @@
+# Mbed TLS Review Guidelines for Contributors
```
davidhorstmann-arm, line 1: Since this already sits under "Mbed TLS > Review" in the docs, it would be good to rename to just "Guidelines for Contributors".
```suggestion
# Guidelines for Contributors
```

reviews/review_guidelines.md
```diff
@@ -1,4 +1,4 @@
-# Mbed TLS Review Guidelines
+# Mbed TLS Review Guidelines for Reviewers
```
davidhorstmann-arm, line 1: Same as above, just "Guidelines for Reviewers" is best.
```suggestion
# Guidelines for Reviewers
```

davidhorstmann-arm (commented):
I have a couple of formatting nits, non-blockers.

reviews/review-for-contributors.md
mpg: Ah, didn't notice, I always use []. Thanks for checking!

reviews/review-for-contributors.md
mpg: > That's not completely true,

Agree, it won't happen all the time, but that fact that it can happen in some circumstances is annoying enough.

> I don't think we should forbid it

Agreed, again I'll adjust the strength of the wording.

mpg:
Ironically, I found myself force-pushing several times in a row because I couldn't seem to get things right for the DCO check. The commit you previously reviewed was 455f9d806030efd96e058d8370f249d36a220190. I believe I have not altered its content, just the commit message.

davidhorstmann-arm (dismissed):
LGTM, thanks!

reviews/review-for-contributors.md
```diff
@@ -12,12 +12,19 @@ See the [guideline for reviewers](<review_guidelines.md>) for a list of things r
 
 ## General considerations
 
-As a project, we find reviewer bandwidth to be our most common bottleneck. As a consequence, the review process should be optimised for reviewers.
+As a project, we find reviewer bandwidth to be our most common bottleneck. As a consequence, the review process should be optimised for reviewers. We recommend the following flow:
 
-Reviewers will generally assign themselves on a PR when there are available to review it, or when they commit to reviewing it later.
+1. Make a (draft) pull request.
+2. For the the CI results.
```
gilles-peskine-arm, line 18: This has a spurious “the” and is missing at least one word.

gilles-peskine-arm (changes requested):
One line isn't quite right, other than that LGTM.

gilles-peskine-arm (approved):
LGTM

PR #94, opened 2023-04-25
Add note to explain angle brackets are included in command

tom-daubney-arm:
The current make command given for compiling with an alternative config file does not work because it doesn't escape characters properly. This PR modifies the command so that it is correct.

### DIFF
--- a/kb/compiling-and-building/how-do-i-configure-mbedtls.md
+++ b/kb/compiling-and-building/how-do-i-configure-mbedtls.md
@@ -22,7 +22,7 @@ The `config.py` script automatically finds the `mbedtls_config.h` file when it r
 
 ## Alternative configuration files
 
-You might want to keep the custom configuration file for your application outside the Mbed TLS source tree. You can do this by defining the macro `MBEDTLS_CONFIG_FILE` for the desired filename (including the quote or angular brackets) at compile time. For example, using **make**:
+You might want to keep the custom configuration file for your application outside the Mbed TLS source tree. You can do this by defining the macro `MBEDTLS_CONFIG_FILE` for the desired filename (including the quote or angular brackets) at compile time. For example, using **make**: (Note: The angle brackets `<>` are included in the command, but they could be replaced with properly escaped double quotes `\"\"`.)
     ```
      CFLAGS="-Ipath/to/config -DMBEDTLS_CONFIG_FILE='<my_config.h>'" make
     ```


paul-elliott-arm (changes requested):

kb/compiling-and-building/how-do-i-configure-mbedtls.md
```diff
@@ -24,7 +24,7 @@ The `config.py` script automatically finds the `mbedtls_config.h` file when it r
 
 You might want to keep the custom configuration file for your application outside the Mbed TLS source tree. You can do this by defining the macro `MBEDTLS_CONFIG_FILE` for the desired filename (including the quote or angular brackets) at compile time. For example, using **make**:
     ```
-     CFLAGS="-Ipath/to/config -DMBEDTLS_CONFIG_FILE='<my_config.h>'" make
+     CFLAGS="-I 'path/to/config' -DMBEDTLS_CONFIG_FILE='\"<my_config.h>\"'" make
```
gilles-peskine-arm, line 27: This change is wrong. It results in being equivalent to
```
#include "<my_config.h>"
```

The original code is correct as long as `path/to/config` doesn't contain special characters such as spaces.

gilles-peskine-arm (changes requested):
Part of the change turns a correct command into a bad one.

kb/compiling-and-building/how-do-i-configure-mbedtls.md
tom-daubney-arm: Little bit confused here. Today I used the following exact command to successfully compile an empty config file:
```CFLAGS="-I '/home/my_user/repos/scratch/mbedtls/configs' -DMBEDTLS_CONFIG_FILE='\"/home/my_user/repos/scratch/mbedtls/configs/empty-config.h\"'" make lib -j```

I am not sure I understand which bit you are saying is wrong?

Comment (gilles-peskine-arm) [reply kb/compiling-and-building/how-do-i-configure-mbedtls.md:27]: That's not the same command. The expansion of the macro `MBEDTLS_CONFIG_FILE` must be either `"filename.h"` or `<filename.h>`, i.e. the file name surrounded either by double quotes or angle brackets. The code in the documentation arranges for angle brackets, which is fine. The code in your latest comment arranges for double quotes, which is fine. The code in your commit puts both double quotes and angle brackets, which doesn't work.

The rest of the quoting are extra layers to get the desired characters through two levels of shell expansion: first the shell that runs make, then the shell that make runs.

Comment (tom-daubney-arm) [reply kb/compiling-and-building/how-do-i-configure-mbedtls.md:27]: Thank you for the explanation. I get the bit about adding the extra layers to get the desired bits through the shell expansion (this is what I thought the original issue was). What I had failed to realise in the original docs is why the angle brackets were there. I thought they were being used in the placeholder manner which I often see in documentation so when I ran the command myself I removed them and just put in the name of my config file and then this didn't work because I didn't have the quotes or the angle brackets.

So, the original docs are fine, I misinterpreted them. Shall I just close this PR?

Comment (davidhorstmann-arm) [reply kb/compiling-and-building/how-do-i-configure-mbedtls.md:27]: I agree this is confusing. I'd be in favour of adding a note to point out that the angle brackets are included here (or switching to using `\"` for the include instead, as it's less likely to be misinterpreted).

Comment (gilles-peskine-arm) [reply kb/compiling-and-building/how-do-i-configure-mbedtls.md:27]: I haven't checked if I was actually the one who wrote this line, but I'd write it with `<>` because they're less of a pain to quote. With double quotes, you have to think about how you're going to quote them due to their effect on the shell.

> I'd be in favour of adding a note to point out that the angle brackets are included here

You mean, something like “including the quote or angular brackets”, in the sentence just above the code snippet?

Comment (davidhorstmann-arm) [reply kb/compiling-and-building/how-do-i-configure-mbedtls.md:27]: > You mean, something like “including the quote or angular brackets”, in the sentence just above the code snippet?

Yes exactly, something like "(note that the angle brackets `<>` are included)".

**Comment by tom-daubney-arm** [comment]
I updated the title of this PR to better reflect it's new purpose and I have pushed a new commit.

File: kb/compiling-and-building/how-do-i-configure-mbedtls.md [hunk 2]
```diff
@@ -22,11 +22,11 @@ The `config.py` script automatically finds the `mbedtls_config.h` file when it r
 
 ## Alternative configuration files
 
-You might want to keep the custom configuration file for your application outside the Mbed TLS source tree. You can do this by defining the macro `MBEDTLS_CONFIG_FILE` for the desired filename (including the quote or angular brackets) at compile time. For example, using **make**:
+You might want to keep the custom configuration file for your application outside the Mbed TLS source tree. You can do this by defining the macro `MBEDTLS_CONFIG_FILE` for the desired filename (including the quote or angular brackets) at compile time. For example, using **make**: (Note: The angle brackets `<>` are included.)
```
gilles-peskine-arm, line 25: Do we really need to say three times that the angle brackets are included???

davidhorstmann-arm: Ah my apologies, I misread [your previous comment](https://github.com/Mbed-TLS/mbedtls-docs/pull/94#discussion_r1177645377). However, I' prefer to keep this note and remove the original one because it's more obviously linked to the code snippet

tom-daubney-arm: There is now one and only one comment about angle brackets in a more obvious place.

davidhorstmann-arm (approved):
LGTM, thanks again!

kb/compiling-and-building/how-do-i-configure-mbedtls.md
```diff
@@ -22,7 +22,7 @@ The `config.py` script automatically finds the `mbedtls_config.h` file when it r
 
 ## Alternative configuration files
 
-You might want to keep the custom configuration file for your application outside the Mbed TLS source tree. You can do this by defining the macro `MBEDTLS_CONFIG_FILE` for the desired filename (including the quote or angular brackets) at compile time. For example, using **make**:
+You might want to keep the custom configuration file for your application outside the Mbed TLS source tree. You can do this by defining the macro `MBEDTLS_CONFIG_FILE` for the desired filename at compile time. For example, using **make**: (Note: The angle brackets `<>` are included.)
```
gilles-peskine-arm, line 25: I know what you mean by “The angle brackets `<>` are included” because I know the context, but it doesn't follow from the meaning of the words. Yes, I can see that there are angle brackets in the line below, so what? The point is that _the reader_ needs to include them (or double quotes) in their command.

gilles-peskine-arm, line 25: Either angle brackets or double quotes are ok: it's what you'd write after an `#include` statement. The current version is missing the information that double quotes are ok.

gilles-peskine-arm (changes requested):
I find the current version misleading and less informative.

kb/compiling-and-building/how-do-i-configure-mbedtls.md
tom-daubney-arm: Yep, fair points. Will figure out a smarter wording for this and resubmit.

tom-daubney-arm:
Please ignore the fact that I closed the PR. I had a slight git meltdown which is now fixed. Apologies.

davidhorstmann-arm (approved):
LGTM, thanks!

paul-elliott-arm (approved):
LGTM too

PR #95, opened 2023-04-26
Update long-term plans: 3.x progress, error codes and more

gilles-peskine-arm:
* I wanted to add the plan to unify `MBEDTLS_ERR_xxx` with `PSA_ERROR_xxx`.
* I also added a note about PQC.
* I also updated things that were done, in progress, or where out plans have changed a bit.

### DIFF
--- a/project/long-term-plans.md
+++ b/project/long-term-plans.md
@@ -44,6 +44,17 @@ Note that the bignum/MPI API, though it's not a crypto API, will also likely be 
 
 The X.509 and TLS APIs (`mbedtls_x509_` and `mbedtls_ssl_` namespaces) will of course remain, only the legacy crypto API is being retired. Some functions in those modules may change signature / argument types, but other than that those APIs are unaffected.
 
+### Simplify and unify error codes
+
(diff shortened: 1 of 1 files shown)


project/long-term-plans.md
```diff
@@ -46,12 +38,23 @@ Currently there are two APIs for most crypto operations:
...
 
 Note that the bignum/MPI API, though it's not a crypto API, will also likely be removed from our public API at this point. This follows the general trend of making more things opaque/private (for example, most struct members became private in Mbed TLS 3.0).
 
 The X.509 and TLS APIs (`mbedtls_x509_` and `mbedtls_ssl_` namespaces) will of course remain, only the legacy crypto API is being retired. Some functions in those modules may change signature / argument types, but other than that those APIs are unaffected.
 
+### Simplify and unify error codes
+
+Currently, `mbedtls_xxx` functions return `MBEDTLS_ERR_xxx` error codes, and `psa_xxx` functions return `PSA_ERROR_xxx` error codes. These use overlapping value spaces, so a function has to pick, it's impossible for a function to return either kind of error code depending on the error.
```
davidhorstmann-arm, line 49: It's unclear what the last part of this means. It could mean either one of:
* It's impossible for a single function to use both different types of error code.
* For certain errors, a function cannot return an error of either type.

I suspect that it is the first option, but could you reword this to make it more obvious?

davidhorstmann-arm (changes requested):
A couple of small things

davidhorstmann-arm (approved):
LGTM, thanks!

PR #98, opened 2023-05-10
Flatten security advisory nesting structure

davidhorstmann-arm:
Remove unnecessary nesting of the security advisories under "Tech Updates". This structure comes from the old website and is confusing within the current documentation.

Redirect the previous page to the new location.


gilles-peskine-arm (approved):
LGTM

PR #99, opened 2023-05-10
Remove unnecessary "Mbed TLS" from Getting Started heading

davidhorstmann-arm:
We know which project's documentation we are in because of the site title. Repeating the project name is redundant and a little jarring.

Shorten the titles to be more intuitive.

### DIFF
--- a/getting_started/index.md
+++ b/getting_started/index.md
@@ -1,4 +1,4 @@
-# Getting started with Mbed TLS
+# Getting Started
 ```{toctree}
 ---
 glob:


tom-cosgrove-arm:
Actually, I like the titles as they are

getting_started/index.md
```diff
@@ -1,4 +1,4 @@
-# Getting started with Mbed TLS
+# Getting Started
```
gilles-peskine-arm, line 1: Like Tom, I don't think having “Mbed TLS” in the title is particularly jarring.

Also, knowledge base article titles currently consistently use normal case, not title case.

davidhorstmann-arm:
Perhaps I can persuade you both with some screenshots from the website. (Remember that the main home of the docs is there rather than GitHub).

In this one we have "Getting started with Mbed TLS" under the main "Mbed TLS" project title. This is particularly odd since "API Reference" and "Project" are assumed implicitly to be relating to Mbed TLS:
getting_started_mbedtls

The "Long-term plans" one is less jarring, admittedly, but it already falls under `Mbed TLS > Project`, so the context of "The Mbed TLS Project" is implied. I'd be happy to compromise and leave this one as-is, seeing as you could argue it's not fully clear what aspect of the project has long-term plans:
project_mbedtls

getting_started/index.md
davidhorstmann-arm: Re title case, this is true of many knowledge base articles, but not of section headings (of which "Getting Started" is now one). Sections are usually title case - for reference, see the section headings: ["Knowledge Base"](https://mbed-tls.readthedocs.io/en/latest/kb/) and ["Compiling and Building"](https://mbed-tls.readthedocs.io/en/latest/kb/compiling-and-building/).

gilles-peskine-arm: Looking further, the “Getting started” page is actually a table of contents page, and there are separate getting-started guides for downloading, for building, and for the PSA API. So I now agree that it makes sense to rename the page to be just “Getting started”.

Regarding title case: I personally dislike title case and think that in the 21st century Internet world it feels affected. But indeed it's currently commonly used for pages that represent sections on the website. In the [old site](https://web.archive.org/web/20190107111130/https://tls.mbed.org/), this was inconsistent (e.g. “Knowledge Base” and “Source Code” but “Dev corner” and “API reference”). Anyway it's a cosmetic thing so I wouldn't reject on that basis.

davidhorstmann-arm: In that case I'll revert the more controversial "Long-term plans" change to its original and just stick with the "Getting Started" rename.

tom-cosgrove-arm:
Given those screenshots, "Project" is the thing that stands out as weird to me!

I might be inclined to suggest "Mbed TLS API Reference", but it seems entirely reasonable for "Long-term plans for Mbed TLS" and "External Reviewers" as they are.  Possibly "Mbed TLS Roadmap" :)

gilles-peskine-arm (approved):
ok

PR #110, opened 2023-07-20
Misc improvements to mtest

daverodgman:
Fix total size calculation
Support disassembly of entire library
Improve docs
Hide size of zero-size modules
Use -Oz when -S specified


tom-cosgrove-arm (approved):
LGTM

PR #113, opened 2023-09-05
mtest - fall back to clang if armclang unavailable for -S

daverodgman:
When running `mtest -S`, fall back to clang if arm clang is not available.

Enable unreachable code warning for clang.


tom-cosgrove-arm (approved):
LGTM

PR #115, opened 2023-09-13
Update armmbed links

daverodgman:
Update links pointing to the old Github org


gilles-peskine-arm (changes requested):
Please also change places where ARMmbed has different case. At least two occurrences of `https://armmbed.github.io/mbed-crypto/html/` which should now point to https://arm-software.github.io/psa-api/crypto/1.1/ .

gilles-peskine-arm (approved):

PR #116, opened 2023-09-27
Misc mtest updates

daverodgman:
Align with TF-M on the CPU for reporting TF-M size
Support new location of TF-M config (changed in https://github.com/Mbed-TLS/mbedtls/pull/8164 )
Improve help
Improve handling of CFLAGS

### DIFF
--- a/tools/bin/mtest
+++ b/tools/bin/mtest
@@ -50,23 +50,24 @@ QEMU_NAME_MAP = {
 }
 
 MEMSAN_OPTIONS = [ "MBEDTLS_TEST_CONSTANT_FLOW_MEMSAN", "-MBEDTLS_USE_PSA_CRYPTO", "-MBEDTLS_AESNI_C", "-MBEDTLS_AESCE_C" ]
-MEMSAN_CFLAGS  = "-fsanitize=memory -O1 -g3 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize-memory-track-origins=2"
+MEMSAN_CFLAGS  = ['-fsanitize=memory', '-O1', '-g3', '-fno-omit-frame-pointer', '-fno-optimize-sibling-calls', '-fsanitize-memory-track-origins=2']
 MEMSAN_CC      = "clang"
 
 COMPILER_CONFIGS = {
 #    "armclang": [ "MBEDTLS_NO_UDBL_DIVISION" ]
 }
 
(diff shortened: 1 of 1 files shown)


tools/bin/mtest
```diff
@@ -238,115 +239,149 @@ def toolchain_exes(target):
...
-            c_flags += " -Oz"
-        else:
-            c_flags += " -Os"
-    if "-std" not in target.cflags: c_flags += " -std=c99"
+    # target_c_flags may get overridden by the user (and dropped from the resulting set of flags);
+    # c_flags is always applied
+
+    target_c_flags = ["-std=c99", f"-I{ROOT}/include"]
+
+    if args.size_tfm and "clang" in target.compiler:
```
tom-cosgrove-arm, line 247: Does this match `armclang`?

tom-cosgrove-arm: Should it use `compiler_family`?

daverodgman: Yes

daverodgman: No benefit to changing to compiler_family: this is maybe poorly named - it means "compiler without version", so "clang" rather than "clang-15", but "armclang" is separate.

tools/bin/mtest
```diff
@@ -397,36 +432,57 @@ def configure_options(target):
...
-            path = config if os.path.exists(config) else f"{ROOT}/configs/{config}"
-            if path.split("/")[-1].startswith("crypto_config_"):
-                dest = CRYPTO_CONFIG_FILE
+        if config == COPY_TFM_CONFIG:
+            # special flag to copy both TF-M config files - these have a couple of possible locations
+            for mc, cc in TFM_CONFIG_FILES:
+                if os.path.exists(f"{ROOT}/configs/{mc}") and (cc is None or os.path.exists(f"{ROOT}/configs/{cc}")):
+                    if mc == TFM_CONFIG_FILES[0][0]:
+                        # latest TF-M config file unsets this, so set it again
+                        set_no_platform_entropy = True
```
tom-cosgrove-arm, line 448: note: #8164 moves the (verbatim) TFM config files to a directory, and has a file `config-tfm.h` that will apply the things like this

tom-cosgrove-arm (approved):
LGTM

tools/bin/mtest
daverodgman: The wrapper disables `MBEDTLS_NO_PLATFORM_ENTROPY`, which we want to set for TF-M size measurement.

PR #117, opened 2023-09-28
Mtest - Disable SPM for TF-M config

daverodgman:
Disable SPM in older versions of the tf-m config file


tom-cosgrove-arm (approved):
LGTM

PR #121, opened 2023-10-06
Mtest update

daverodgman:
- Improve setting up compiler flags, especially when using `-S`
- Automatically set some TF-M options that are not yet in their config file


tom-cosgrove-arm (approved):
LGTM

PR #133, opened 2024-02-01
Bring over remaining phabricator content

daverodgman:
Port remaining phabricator wiki content:

- security centre (ie vulnerability reporting process)
- notes on testing

I've lightly updated the testing notes - there may still be some out-of-date info but I've tried to fix all the major things. I've removed some things that tend to go out of date (e.g. list of supported tool versions).

Non-critical out-of-date things can be fixed in a follow-up and shouldn't scope-creep this PR.

Part of review: please confirm that I haven't missed any useful content from Phabricator https://developer.trustedfirmware.org/w/mbed-tls/

### DIFF
--- a/kb/testing/testing-ci.md
+++ b/kb/testing/testing-ci.md
@@ -104,4 +104,4 @@ The tests attempt to establish an IPv6 connection to the local host. This is opt
 
 ### Docker environment
 
-A Dockerfile is provided in [tests/docker](https://github.com/ARMmbed/mbedtls/tree/development/tests/docker).
+A Dockerfile is provided in [tests/docker](https://github.com/Mbed-TLS/mbedtls/tree/development/tests/docker).
--- a/kb/testing/testing-ci.md
+++ b/kb/testing/testing-ci.md
@@ -101,7 +101,3 @@ ln -s x86_64-linux-gnu/asm /usr/include/asm
 ### Notes about networking
 
 The tests attempt to establish an IPv6 connection to the local host. This is optional, and will be skipped if IPv6 is not available. However, if IPv6 is possible but IPv6 connections to localhost are blocked by a firewall, the test will fail.
-
-### Docker environment
-
-A Dockerfile is provided in [tests/docker](https://github.com/Mbed-TLS/mbedtls/tree/development/tests/docker).
--- a/kb/testing/testing-ci.md
+++ b/kb/testing/testing-ci.md
@@ -1,0 +1,107 @@
+# Mbed TLS CI
+
+All code that is included in Mbed TLS must be contributed in the form of [[https://github.com/Mbed-TLS/mbedtls/pulls | pull requests on GitHub]] and undergoes some automated testing. This page describes the continuous integration jobs that run on every pull request.
+
+## DCO
+
+This job checks that all commits have a `Signed-off-by:` line. The presence of this line indicates that the author of the commit certifies that the commit is covered by the [[https://github.com/Mbed-TLS/mbedtls/blob/development/dco.txt | Developer Certificate of Origin]] and contributed according to the [[https://github.com/Mbed-TLS/mbedtls/blob/development/README.md#License | project license]].
+
+All commits must have such a line, otherwise the commit cannot be accepted for legal reasons. As a temporary exception, commits from Arm employees or from other contributors who already have a contributor license agreement (CLA) can still be accepted, but please include a `Signed-off-by:` line in any new work.
+
+If the DCO job fails, please reword all commit messages that are missing a `Signed-off-by:` line. If you have multiple commit messages to rewrite, [[https://stackoverflow.com/questions/25570947/how-to-use-git-interactive-rebase-for-signing-off-a-series-of-commits | How to use git interactive rebase for signing off a series of commits]] may help.
+
+## PR-head and PR-merge jobs
+
+The PR-//NNN//-head and PR-//NNN//-merge jobs run an extensive battery of tests on several platforms. The -head jobs run the tests on the tip of the submitted code. The -merge jobs run the tests on a merge with the target branch.
+
+The PR-head job runs in public-facing CI every time a pull request is updated.
+
+### High-level overview of the Jenkins test coverage
+
+The Jenkins PR job includes the following parts:
+
+- Run `tests/scripts/all.sh` on Ubuntu 16.04 x86 (64-bit). This script includes:
+    - Some sanity checks.
+    - Tests of the library in various configurations. The tests are mainly unit tests (`make test`), SSL feature tests (`tests/ssl-opt.sh`) and interoperability tests (`tests/compat.sh`).
+    - Some cross compilation with GCC-arm, Arm Compiler 5 (`armcc`), Arm Compiler 6 (`armclang`) and MinGW (`i686-w64-mingw32-gcc`). These are only builds, not tests.
+- Run a subset of `tests/scripts/all.sh` on FreeBSD (amd64)
+- Build on Windows with MinGW and Visual Studio. We use the following Visual Studio versions:
+    - Since Mbed TLS 2.19: VS 2013, 2015, 2017. As of January 2022, we expect to drop VS 2013 soon.
+
+The component names are:
+
+* all_sh-//platform//-//component//: `tests/scripts/all.sh -k `//component// run on //platform//.
+* `code-coverage`: `tests/scripts/basic-build-test.sh` run on Linux (currently Ubuntu 16.04).
+* `win32-mingw`, `win32_msvc12_32`, `win32-msvc12_64`: ad hoc Windows jobs in the default configuration, see below.
+* `Windows-`//20NN//: builds with the specified version of Visual Studio (Visual Studio 10 2010, Visual Studio 12 2013, Visual Studio 14 2015 or Visual Studio 15 2017).
+    * Library configuration: `config.py full` (`config.pl full` in older branches) minus `MBEDTLS_THREADING_xxx`.
+    * Visual Studio configurations: `Release` and `Debug`.
+    * Architectures: `Win32` and `x64` (except no `x64` in VS 2010).
+    * `PlatformToolset` property:
+        * “Retargeted” build: set to the version corresponding to the Visual Studio version for “retargeted” builds.
+        * “Non-retargeted” build: set to `v120` on `development` (C99 code base).
+    * Testing: `selftest.exe`, plus the unit tests in the CMake builds.
+
+The `win32-mingw` component runs the following bat script:
+```
+cmake . -G "MinGW Makefiles" -DCMAKE_C_COMPILER="gcc"
+mingw32-make
+mingw32-make test
+ctest -VV
+programs\test\selftest.exe
+```
+
+The `win32_msvc12_32` component runs the following bat script:
+```
+call "C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat"
+cmake . -G "Visual Studio 12"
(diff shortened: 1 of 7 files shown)


kb/testing/testing-ci.md
```diff
@@ -0,0 +1,107 @@
...
+
+The Jenkins PR job includes the following parts:
+
+- Run `tests/scripts/all.sh` on Ubuntu 16.04 x86 (64-bit). This script includes:
+    - Some sanity checks.
+    - Tests of the library in various configurations. The tests are mainly unit tests (`make test`), SSL feature tests (`tests/ssl-opt.sh`) and interoperability tests (`tests/compat.sh`).
+    - Some cross compilation with GCC-arm, Arm Compiler 5 (`armcc`), Arm Compiler 6 (`armclang`) and MinGW (`i686-w64-mingw32-gcc`). These are only builds, not tests.
+- Run a subset of `tests/scripts/all.sh` on FreeBSD (amd64)
+- Build on Windows with MinGW and Visual Studio. We use the following Visual Studio versions:
+    - Since Mbed TLS 2.19: VS 2013, 2015, 2017. As of January 2022, we expect to drop VS 2013 soon.
```
tom-cosgrove-arm, line 29: January 2022?

gilles-peskine-arm:
This resolves the content part of https://github.com/Mbed-TLS/mbedtls/issues/4086 (with a remaining task of hunting down and updating links to the old wiki).

kb/testing/testing-constant-flow.md
```diff
@@ -0,0 +1,99 @@
...
+- Often provides some guarantee of safety
+- Can sometimes transform more readable unsafe code into safe code.
+
+Cons:
+- May involve rewriting sections of constant-time code
+- Still potential for safe code but unsafe binaries
+
+## Empirical methods
+
+There appears to be some empirical testing tools available for testing constant time code. Tools such as [[ https://github.com/oreparaz/dudect#dudect-dude-is-my-code-constant-time | dudect ]] work by running the program with different inputs and trying to detect statistically different execution times. While these tools may not provide formal guarantees of safety against side-channel attacks it could be a useful, and easy to implement, metric for constant time code.
```
tom-cosgrove-arm: s/appears/appear/ since mre than one tool (and agrees with "there")

gilles-peskine-arm:
> I've removed some things that tend to go out of date (e.g. list of supported tool versions).

Please do keep the list of tool versions, that is the main information in https://developer.trustedfirmware.org/w/mbed-tls/testing/ci/. It documents the tool versions that we have in our CI, for which otherwise you have to dig down in multiple CI scripts. It's a checklist that team members use to set up their environment when they join.

kb/testing/testing-ci.md
```diff
@@ -0,0 +1,107 @@
+# Mbed TLS CI
+
+All code that is included in Mbed TLS must be contributed in the form of [[https://github.com/Mbed-TLS/mbedtls/pulls | pull requests on GitHub]] and undergoes some automated testing. This page describes the continuous integration jobs that run on every pull request.
+
+## DCO
+
+This job checks that all commits have a `Signed-off-by:` line. The presence of this line indicates that the author of the commit certifies that the commit is covered by the [[https://github.com/Mbed-TLS/mbedtls/blob/development/dco.txt | Developer Certificate of Origin]] and contributed according to the [[https://github.com/Mbed-TLS/mbedtls/blob/development/README.md#License | project license]].
```
tom-cosgrove-arm: I'd not come across [[url | text]] for links in markdown before - we seem to normally use [text](url) - and that form is used later in this document, too. Do we care enough about consistency to change one or the other?

daverodgman:
> > I've removed some things that tend to go out of date (e.g. list of supported tool versions).
> 
> Please do keep the list of tool versions, that is the main information in https://developer.trustedfirmware.org/w/mbed-tls/testing/ci/. It documents the tool versions that we have in our CI, for which otherwise you have to dig down in multiple CI scripts. It's a checklist that team members use to set up their environment when they join.

I don't agree, that list was last updated in Sept 2020, and I doubt we would remember to update it in the future. So I don't think it's providing useful information.

kb/testing/testing-ci.md
daverodgman: Oh, that's phabricator syntax - I updated most but obviously missed some. I'll fix it.

kb/testing/testing-ci.md
```diff
@@ -0,0 +1,107 @@
...
+ln -s x86_64-linux-gnu/asm /usr/include/asm
+```
+
+### Notes about networking
+
+The tests attempt to establish an IPv6 connection to the local host. This is optional, and will be skipped if IPv6 is not available. However, if IPv6 is possible but IPv6 connections to localhost are blocked by a firewall, the test will fail.
+
+### Docker environment
+
+A Dockerfile is provided in [tests/docker](https://github.com/Mbed-TLS/mbedtls/tree/development/tests/docker).
```
tom-cosgrove-arm, line 107: I wonder if we should keep some of the warning from the original along the lines of "This Dockerfile is not identical to the one used in the CI (XXX is it?) and is infrequently updated" or something like that.

How does this one compare to the one at (for example) https://github.com/Mbed-TLS/mbedtls-test/blob/master/resources/docker_files/ubuntu-18.04/Dockerfile ? (and 16.04, 20.04, 22.04)

Would we be better off removing this one and referring to the other(s)?

daverodgman: I don't know enough about how we use docker to know how best to document this. If you want to suggest something better I'm happy to take it.

gilles-peskine-arm: https://github.com/Mbed-TLS/mbedtls-test/blob/master/resources/docker_files/ubuntu-18.04/Dockerfile is not infrequently updated, it's unmaintained and to be removed. Part of the work of removing it is to remove references to it in various documentation, so please do not reference it in any new document.

daverodgman: I've removed the section; I'll leave it to someone who knows more about our docker usage to add an updated version in a separate PR. (@bensze01  ?)

tom-cosgrove-arm: @gilles-peskine-arm The various Dockerfiles in https://github.com/Mbed-TLS/mbedtls-test/blob/master/resources/docker_files/ubuntu-??.04/Dockerfile were updated several times in 2023, including by you in September. Are we really removing those? I've found them useful

If you mean the one at https://github.com/Mbed-TLS/mbedtls/tree/development/tests/docker referenced in this text, then yes, it seems unmaintained, and we shouldn't refer to it

gilles-peskine-arm: Oops, right, I copied the wrong link. I meant the one referenced in this text, which was an external contribution back when the ones in `mbedtls-test` weren't public. We have no plan to do anything other than routine maintenance to the dockerfiles in `mbedtls-test`.

kb/testing/testing-constant-flow.md
```diff
@@ -0,0 +1,99 @@
...
+- https://github.com/llvm-mirror/compiler-rt/blob/master/include/sanitizer/msan_interface.h
+
+Pros:
+- clang/memsan is already used in our CI
+- much faster than valgrind/memcheck (only 2-3x slowdown, so about 10x faster than using valgrind)
+
+Cons:
+- only works with clang on Linux, FreeBSD, NetBSD
+- being compile-time instrumentation, limited with respect to inline assembly
+- for the same reason, misses branches in toolchain-provided functions (for example long multiplication, see [this paper](https://core.ac.uk/download/pdf/188281541.pdf)) which valgrind would see
```
tom-daubney-arm, line 67: The link to the paper is broken here and I am afraid I am not familiar with which paper is supposed to be linked here.

tom-daubney-arm (changes requested):
Very nearly fine IMO. With one exception the links all work as expected and I am satisfied that nothing major has been left out.

Please see my comment for the broken link. I was going to find you the working link but I don't know where the paper can be found.

mpg:
> It documents the tool versions that we have in our CI, for which otherwise you have to dig down in multiple CI scripts.

Multiple CI scripts really? My understanding is that you only have to look at a single file: https://github.com/Mbed-TLS/mbedtls-test/blob/master/resources/docker_files/ubuntu-22.04/Dockerfile (or whatever version's closest to the distro/version you're using). I has all the info that the table has I think - except as Dave points out, the Dockefiles are always up-to-date while the table isn't. (Case in point: when we added named versions of GnuTLS & OpenSSL on the CI, we didn't update the table.)

I think we should replace the table with a link to one of the Dockerfiles. If there's any additional information needed, we can keep it _in addition_ to the link. For example draw attention to the fact that you need to set up environment variables pointing to GnuTLS & OpenSSL, talk about which versions of Python are supported, because in the Dockerfiles that's mostly implicit (and even then, perhaps we shouldn't directly mention versions of Python here but instead point to the README from the main repo - that's what's most likely to remain up-to-date).

gilles-peskine-arm:
> Multiple CI scripts really? My understanding is that you only have to look at a single file

You have to look at all the docker files. And `all.sh` to understand how some of the tools are invoked. And `scripts/min_requirements.py` (from which you deduce `scripts/min_requirements.py`) which is invoked by Groovy code. Our scripts have a level of complexity where “TRFS, noob” doesn't cut it. Remember, this is information someone needs when they're starting working on the project.

> the Dockefiles are always up-to-date while the table isn't.

Sure, that's the curse of documentation. It's still better to have documentation. Documentation in the canonical place rather than a wiki that's being shut down would have a better chance of keeping up-to-date.

daverodgman:
I agree with Manuel. History has shown that we will not keep it up to date. I also think it's low-value information - most people will never need it and can always ask. A short section to document "here is how you find the minimum tool versions from docker" is fine.

I'm not going to do this in this PR, but if anyone feels motivated, happy to take a separate PR to add such a section.

kb/testing/testing-constant-flow.md
daverodgman: Searching core.ac.uk for that number yields https://core.ac.uk/search?q=188281541 which looks like it matches the description here. @mpg does that look like the right link to you?

tom-cosgrove-arm: I am not Manuel, but this certainly seems to match. We should probably cite papers properly, rather than just providing links that may go stale (or at least ask the Wayback machine to archive them!)

daverodgman: I've linked it properly.

tom-daubney-arm (dismissed):
LGTM - thanks for fixing the link

kb/testing/testing-constant-flow.md
```diff
@@ -0,0 +1,101 @@
...
+- Often provides some guarantee of safety
+- Can sometimes transform more readable unsafe code into safe code.
+
+Cons:
+- May involve rewriting sections of constant-time code
+- Still potential for safe code but unsafe binaries
+
+## Empirical methods
+
+There appears to be some empirical testing tools available for testing constant time code. Tools such as [dudect](https://github.com/oreparaz/dudect#dudect-dude-is-my-code-constant-time) work by running the program with different inputs and trying to detect statistically different execution times. While these tools may not provide formal guarantees of safety against side-channel attacks it could be a useful, and easy to implement, metric for constant time code.
```
tom-cosgrove-arm, line 97: My comment from yesterday `s/appears/appear/` still stands "There appear to be some empirical testing tools available" reads better

daverodgman: Fixed, but for future reference, this level of typo isn't worth asking for more iterations

tom-cosgrove-arm (approved):
LGTM

PR #136, opened 2024-02-13
Add script to be able to kick the CI to run on a specified PR

tom-cosgrove-arm:


tools/bin/mbedtls-kick-ci
```diff
@@ -0,0 +1,100 @@
+#!/bin/sh
+
+set -eu
+
+script=$(basename $0)
```
gilles-peskine-arm, line 5:
```suggestion
script=$(basename -- "$0")
```
or
```
script=${0##*/}
```

tools/bin/mbedtls-kick-ci
```diff
@@ -0,0 +1,100 @@
...
+# user ID. Both internal and OpenCI user names will almost certainly be the same
+# as your GitHub ID.
+#
+# To get an API token, choose Configure from your user page, click Add new Token,
+# give it a name, then click Generate. You get one chance to copy it, so do it
+# now! Then put it into ~/bin/jmt.
+#
+# Make sure ~/bin/jmt is chmod 0600.
+
+. ~/bin/jmt
```
gilles-peskine-arm, line 29: Please make this path configurable or make it a dot file.

tools/bin/mbedtls-kick-ci
```diff
@@ -0,0 +1,100 @@
...
+            echo " - FAILED: \\c"
+            echo "$OUTPUT" | perl -n -e 'print "$1\n" if /<title>(.*?)<.title>/;'
+            exit 2
+        fi
+    fi
+    did_something=1
+}
+
+if [ $# = 1 ]; then
+    set -- $1 all
```
gilles-peskine-arm, line 55:
```suggestion
    set -- "$1" all
```

tools/bin/mbedtls-kick-ci
```diff
@@ -0,0 +1,100 @@
...
+        else
+            echo " - FAILED: \\c"
+            echo "$OUTPUT" | perl -n -e 'print "$1\n" if /<title>(.*?)<.title>/;'
+            exit 2
+        fi
+    fi
+    did_something=1
+}
+
+if [ $# = 1 ]; then
```
gilles-peskine-arm, line 54: Technically this works, but it's a bit jarring: `=` is string equality, but it's ok because `$#` is guaranteed to be represented in decimal. The semantically correct way would be
```suggestion
if [ $# -eq 1 ]; then
```

tools/bin/mbedtls-kick-ci
```diff
@@ -0,0 +1,100 @@
...
+    echo $usage 1>&2
+    exit 255
+fi
+
+PR="$1"
+SCOPE="$2"
+
+case "$PR" in
+    [0-9][0-9][0-9][0-9]) : ;;          # all Mbed TLS PRs are four digits for now
+    *) echo $usage 1>&2; exit 255 ;;    # doesn't look like a number
```
gilles-peskine-arm, line 68: We're already past 9000. Please accept variable-length numbers.

Also, please make the error message explicit.

Also, by convention, statuses 126 and above are reserved (can't execute, signal).

```suggestion
    *[!0-9]*) echo >&2 "$0: Fatal: PR number is not a number: $1"; exit 120;;
```

tools/bin/mbedtls-kick-ci
```diff
@@ -0,0 +1,100 @@
...
+
+case $SCOPE in
+    all|openci|openci-pr-merge)
+        kick_one openci-pr-merge $JENKINS_OPENCI_USER_ID $JENKINS_OPENCI_API_TOKEN "https://$host_open_ci/job/mbed-tls-pr-merge/view/change-requests/job/PR-$PR-merge/build/api/json?delay=0sec"
+        ;;
+esac
+
+if [ "$did_something" != 1 ]; then
+    echo $usage 1>&2
+    exit 255
```
gilles-peskine-arm, line 99:
```suggestion
    echo >&2 "$0: Fatal: scope not recognized: $SCOPE"
    exit 2
```

gilles-peskine-arm (changes requested):
Thanks for sharing this script! I've used it successfully after editing the `~/bin/jmt` line.

tools/bin/mbedtls-kick-ci
```diff
@@ -0,0 +1,100 @@
...
+SCOPE="$2"
+
+case "$PR" in
+    [0-9][0-9][0-9][0-9]) : ;;          # all Mbed TLS PRs are four digits for now
+    *) echo $usage 1>&2; exit 255 ;;    # doesn't look like a number
+esac
+
+did_something=0
+
+case "$SCOPE" in
```
gilles-peskine-arm, line 73: Feature request: I'd like to be able to pass multiple job names, typically
```
mbedtls-kick-ci 1234 openci-pr-head openci-pr-merge
```

gilles-peskine-arm:
As discussed, I've pushed an update to use a more standard location for the configuration file.

My other comments are non-blockers. I might handle them in a follow-up PR but I don't intend to push further updates here.

tools/bin/mbedtls-kick-ci
gilles-peskine-arm: Done in https://github.com/Mbed-TLS/mbedtls-docs/pull/136/commits/8953ced81a670a01d4ddf5edb4c00a8e59903857, with the path now under `~/.config`. It can of course still be a symbolic link.

tom-cosgrove-arm:
@gilles-peskine-arm Thanks for the config commit. I've fixed most of the other small niggles.

To whoever reviews: approval from me on Gilles' commit

tools/bin/mbedtls-kick-ci
tom-cosgrove-arm: I've changed the exit code, but "scope not recognised" means "usage was wrong", and scope is an internal thing not mentioned on the command line, so I've left the error message

tools/bin/mbedtls-kick-ci
gilles-peskine-arm: `mbedtls-kick-ci 1234 openci` covers that particular example, actually. So no big deal.

gilles-peskine-arm (approved):
LGTM

PR #137, opened 2024-02-15
Mtest - minor bugfixes

daverodgman:
Minor bugfixes, and fix a Python 3.7 issue

### DIFF
--- a/tools/bin/mtest
+++ b/tools/bin/mtest
@@ -894,9 +894,10 @@ Build targets: build the library by default, or files (e.g. aes.o), programs (e.
     options_with_comma = [x for x in args.config_options if "," in x]
     if options_with_comma:
         # extract raw list for generating cross-product (discard leading hyphens)
-        options_with_comma = set(x[1:] if x.startswith("-") else x for x in ",".join(options_with_comma).split(','))
+        options_with_comma = [x[1:] if x.startswith("-") else x for x in ",".join(options_with_comma).split(',')]
+        options_with_comma = list(dict.fromkeys(options_with_comma)) # remove duplicates
         if "" in options_with_comma: options_with_comma.remove("")
-        args.config_options     = [x for x in args.config_options if "," not in x]
+        args.config_options = [x for x in args.config_options if "," not in x]
         # generate cross-product
         dash_combinations = itertools.product(["", "-"], repeat=len(options_with_comma))
         configs_xprod = [[dash + option for dash, option in zip(dashes, options_with_comma)] for dashes in dash_combinations]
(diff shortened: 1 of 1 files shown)


tools/bin/mtest
```diff
@@ -502,19 +507,23 @@ def get_crypto_config_file():
...
     if option.startswith("PSA_WANT"):
         backup_config_options(crypto_config_path)
         if not silent:
-            log(f"scripts/config.py -f {crypto_config_path} {action} {option} {'1' if action == 'set' else ''}")
-        if action == "set":
-            output, success = subrun([f"scripts/config.py"] + force + ["--file", crypto_config_path, action, option, "1"], silent=True, ignore_failure=True)
-        else:
-            output, success = subrun([f"scripts/config.py"] + force + ["--file", crypto_config_path, action, option], silent=True, ignore_failure=True)
+            log(f"scripts/config.py -f {crypto_config_path} {action} {option} {''.join(value)}")
+        output, success = subrun([f"scripts/config.py"] + force + ["--file", crypto_config_path, action, option] + value, silent=True, ignore_failure=True)
```
minosgalanakis, line 521: If force value are meant to be single element structures, we could use them as such str, int, and then convert in lists on the subrun

~~~~~~
output, success = subrun([f"scripts/config.py"] + [force] + ["--file", crypto_config_path, action, option] + [value], silent=True,
~~~~~~~

tools/bin/mtest
```diff
@@ -502,19 +507,23 @@ def get_crypto_config_file():
 
 def set_config_option(action, option, crypto_config_path, silent=True):
     force = [ "--force" ] if action == "set" else []
```
minosgalanakis, line 509: Would be more readable if it was written like:

~~~~~
force = [] 
    if action == "set":
        force.append("--force")
        if "=" in option:
~~~~~

tools/bin/mtest
```diff
@@ -888,8 +897,10 @@ Build targets: build the library by default, or files (e.g. aes.o), programs (e.
     options_with_comma = [x for x in args.config_options if "," in x]
     if options_with_comma:
         # extract raw list for generating cross-product (discard leading hyphens)
-        options_with_comma = set(x[1:] if x.startswith("-") else x for x in ",".join(options_with_comma).split(','))
-        args.config_options     = [x for x in args.config_options if "," not in x]
+        options_with_comma = [x[1:] if x.startswith("-") else x for x in ",".join(options_with_comma).split(',')]
+        options_with_comma = list(dict.fromkeys(options_with_comma)) # remove duplicates
```
minosgalanakis, line 901: That can be achieved by sets?

minosgalanakis (approved):
LGTM

PR #138, opened 2024-02-27
mtest: automatically set -W for old versions that need it

daverodgman:
As per request from @tom-cosgrove-arm


tom-cosgrove-arm (approved):
LGTM

PR #143, opened 2024-04-23
Archive-Epic: Added script to help moving project items.

minosgalanakis:
This this a simple utility script that allows migration of projects from the [Epics Board](https://github.com/orgs/Mbed-TLS/projects/1) to the [Past Epics](https://github.com/orgs/Mbed-TLS/projects/13/views/1).

It wraps aroud the [github cli ](https://cli.github.com/) so it requires a valid token.  In some cases it has added complexity because the newly introduced [gh project api](https://cli.github.com/manual/gh_project) is inconsistent on the argument requirements, and had to manually use graphQL querries to extract the information.

### DIFF
--- a/tools/bin/mbedtls-archive-epic.py
+++ b/tools/bin/mbedtls-archive-epic.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python3
+""" Assist with the transition of a projectv1 to a projectv2. 
+    This is currently used for management purposes and archiving EPICS.
+
+    It will only move public issues/pr's that have been closed/merged.
+
+    Requires a fully woking gh cli https://cli.github.com/ set-up
+    and an access token with the appropriate permissions.    
+"""
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License") you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import sys
+import json
+import requests
+from shlex import split
+from string import Template
+from subprocess import Popen, check_output, PIPE
+from pprint import pprint
+
+token = os.environ['GITHUB_TOKEN']
+
+
+######################### User Config ###########################
+# Instuctions
+# - Please ensure that your github cli token is present in your enviroment
+# - Set the epic name to tranfer here.
+# - Create the epic column to the arhive board, matching the src name.
+# - If needed adjust the closed_after date to be in a resonable window
+# - Fine tune max entries selection if some entries are missing.
+# - Private issues and drafts are not picked up by the script.
+
+epic = "Mbed TLS 3.6 release"       # Name of epic that needs to be moved
+                                    # (It needs to match an empty epic on v2 board)
+closed_after_date = "2023-10-01"    # Only include entries closed after that date
+
+max_entries = 500                   # How many maximum entries the backend will populate.
+debug = False                       # Set to True to print out the github cli commands.
+
+############################# Constants #############################
+"""
+The following values should remain the same across the project.
+If need to modify them, they can querried as follows:
+
+gh project list --owner Mbed-TLS number 13
+    NUMBER  TITLE                      STATE  ID
+    13      Past EPICs                 open   PVT_kwDOBcuPHc4AgPo7
+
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
Does this still work after the GitHub board migration?

minosgalanakis:
I have not tried it to be honest, but we will definately need to retrofit it, since it was moving inbetween legacy to v2 type of projects. 

Ideally we could replace it with something using pygithub, and it can be done in phases, since it was hacked around in a flat structure where each function does not depend to the others.

tools/bin/mbedtls-archive-epic.py
```diff
@@ -0,0 +1,340 @@
...
+import os
+import sys
+import json
+import requests
+from shlex import split
+from string import Template
+from subprocess import Popen, check_output, PIPE
+from pprint import pprint
+
+token = os.environ['GITHUB_TOKEN']
```
gilles-peskine-arm, line 34: Please support `gh auth token` if the environment variable is unset.

Also please arrange for `mbedtls-archive-epic.py --help` to work even when no token is available.

tools/bin/mbedtls-archive-epic.py
```diff
@@ -0,0 +1,340 @@
...
+# - If needed adjust the closed_after date to be in a resonable window
+# - Fine tune max entries selection if some entries are missing.
+# - Private issues and drafts are not picked up by the script.
+
+epic = "3.6.1 patch release"       # Name of epic that needs to be moved
+                                    # (It needs to match an empty epic on v2 board)
+closed_after_date = "2024-01-01"    # Only include entries closed after that date
+
+max_entries = 500                   # How many maximum entries the backend will populate.
+debug = True                       # Set to True to print out the github cli commands.
```
gilles-peskine-arm, line 51: Can these please be command line items? I don't want to have to modify the script to make it do something useful. Also, because there are prefilled default, I'm afraid that I'd forget to change something and I'd end up modifying something I didn't intend.

tools/bin/mbedtls-archive-epic.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+
+    def __init__(self, success, stdout, stderr):
+
+        self.success = success
+        self.stdout = stdout
+        self.stderr = stderr
+
+
+def do_shell_exec(exec_string, expected_result=0):
```
gilles-peskine-arm, line 117: Shouldn't this just be `subprocess.check_output`?

* I don't see this being used for the error cases, except to print additional error messages that aren't particularly helpful.
* Sometimes the caller forgets to check the return code, it would be better to keep using an exception on failure.
* Callers don't actually rely on having a shell, they're just passing arguments to one program, sometimes with dodgy quoting. So it would be easier to not use a shell.

tools/bin/mbedtls-archive-epic.py
```diff
@@ -0,0 +1,340 @@
...
+        print("Error: Invalid entry for status:", status)
+        return
+    is_closed = "closed" == status
+    issues = get_issues_prs_for_project(issues=True, is_closed=is_closed, repo=repo)
+    output["issues"] = link_issues_pr_linked_to_epic(issues, column_name, projectboard)
+
+    prs = get_issues_prs_for_project(issues=False, is_closed=is_closed, repo=repo)
+    output["prs"] = link_issues_pr_linked_to_epic(prs, column_name, projectboard)
+
+    # If both is selected just toggle the is_closed state and append the results
```
gilles-peskine-arm, line 216: That seems a bit convoluted. Why not just omit both `is:open` and `is:closed` when `status=="both"`?

tools/bin/mbedtls-archive-epic.py
```diff
@@ -0,0 +1,340 @@
...
+                        linked.append(k)
+    return linked
+
+def get_epic_items(column_name, status="both", projectboard=active_project, repo=projects_repo):
+    """ Return a structured data-set containing an epic name and all the open/closed issues in it."""
+
+    output = {"epic": column_name,
+              "issues": [], "prs": []}
+
+    if status not in ["open", "closed", "both"]:
```
gilles-peskine-arm, line 206: Should be an enum?

Or just not bother, since this function is always called with `status="both"`?

tools/bin/mbedtls-archive-epic.py
```diff
@@ -0,0 +1,340 @@
+#!/usr/bin/env python3
+""" Assist with the transition of a projectv1 to a projectv2.
```
gilles-peskine-arm, line 2: Or between projectv2, surely? AFAICT this does work FROM v2 as well, and it's what we want going forward.

gilles-peskine-arm (changes requested):
Thanks for sharing this! I've done a superficial review, and this mostly looks sensible to me, but there are two pain points in making this script accessible to people other than its author:

* It's hard to get started with the script due to hard-coded assumptions.
* Some parts (error handling, string escaping) seem fragile, so I don't know when I can trust the script and when I should be careful.

On the flip side, thanks for the documentation! Especially “User Config” and the explanation of how to find the magic strings `project_xxx`.

PR #145, opened 2024-05-02
Add script to move files to the framework

davidhorstmann-arm:
Pushing a small script that moves files to the `framework` repo from Mbed TLS, while retaining history.

This script works but may not be in a fully-polished state.

### DIFF
--- a/tools/bin/mbedtls-move-to-framework
+++ b/tools/bin/mbedtls-move-to-framework
@@ -1,0 +1,212 @@
+#!/usr/bin/env python3
+"""Move files from the Mbed TLS repository to the framework repository.
+
+Invoke this script from a Git working directory with a branch of Mbed TLS.
+"""
+
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import os
+import pathlib
+import re
+import subprocess
+import sys
+from typing import List, Optional, Tuple, Dict
+
+
+class RepoFileMover:
+    """Move files between repos while retaining history."""
+
+    GIT_EXE = 'git'
+    DEFAULT_BRANCH = 'main'
+
+    def __init__(self, source_repo: str, dest_repo: str,
+                 source_branch_name: str, dest_branch_name: str,
+                 file_map: Dict[str, str]):
+        self._source_repo = os.path.abspath(source_repo)
+        self._dest_repo = os.path.abspath(dest_repo)
+
+        self._src_branch_name = source_branch_name
+        self._tmp_framework_branch_name = 'tmp-branch-move-files-to-framework'
+        self._framework_branch_name = dest_branch_name
+
+        self._file_map = {os.path.normpath(k): os.path.normpath(v) for k, v in file_map.items()}
+
+    def run_git(self, git_cmd: List[str], **kwargs) -> str:
+        """Run a git command and return its output."""
+        if 'universal_newlines' not in kwargs:
+            kwargs['universal_newlines'] = True
+        cmd = [self.GIT_EXE] + git_cmd
+        return subprocess.check_output(cmd, **kwargs)
+
+    def add_file_move(self, src_path: str, dst_path: str) -> None:
+        """Move file at relative path src_path in the source repo to dst_path
+        in the destination repo"""
+        self._file_map[src_path] = dst_path
+
+    def ensure_dirs_exist(self, path: str) -> None:
+        """Make sure all of the required directories exist for
+        moving a given file or directory to it."""
+
+        # Remove any trailing slashes
+        while path[-1] == os.sep:
+            path = path[:-1]
+
+        # Get the directory names
+        dirs, file = os.path.split(path)
+        # Make all required directories
+        os.makedirs(dirs, exist_ok=True)
+
+    def create_dest_repo_branch(self):
+        """Create the branch containing the moved files only"""
+
+        # Create a new branch
+        self.run_git(['checkout', '-b', self._tmp_framework_branch_name])
+
+        # Get a list of all files in the repo
+        repo_files = self.run_git(['ls-files']).split()
+
+        # Get a set of the files we are moving (taking account of directories)
+        files_to_retain = set()
+        for f in self._file_map:
+            if os.path.isdir(f):
+                dir_files = self.run_git(['ls-files', f]).split()
+                dir_files = [os.path.normpath(dir_file) for dir_file in dir_files]
+                files_to_retain.update(dir_files)
+            else:
+                files_to_retain.add(os.path.normpath(f))
+
+        # Delete all files except the ones we are moving
+        for f in repo_files:
+            if os.path.normpath(f) not in files_to_retain:
+                self.run_git(['rm', f])
+
+        # Rename files as requested
+        for f in self._file_map:
+            # Git won't let us move things to themselves
+            if self._file_map[f] != f:
+                self.ensure_dirs_exist(self._file_map[f])
+                self.run_git(['mv', f, self._file_map[f]])
+
+        # Commit the result
+        commit_message = "Move some files to framework repository"
+        self.run_git(['commit', '-asm', commit_message])
+
+    def create_src_repo_branch(self):
+        """Create the branch deleting the moved files"""
+
+        # Create a new branch
+        self.run_git(['checkout', '-b', self._src_branch_name])
+
+        # Delete the moved files
+        files_to_delete = self._file_map.keys()
+        for f in files_to_delete:
+            if os.path.isdir(f):
+                self.run_git(['rm', '-r', f])
+            else:
+                self.run_git(['rm', f])
+
+        # Commit the result
+        commit_message = "Move some files to framework repository"
+        self.run_git(['commit', '-asm', commit_message])
+
+    def merge_files_into_framework(self):
+        """Add the source repo as a remote, pull in the destination branch
+        and merge into a new branch"""
+        # Change to the desination repo
+        os.chdir(self._dest_repo)
+
+        # Fetch/checkout the branch that has the moved files on it
+        self.run_git(['fetch', self._source_repo,
+                      self._tmp_framework_branch_name + ':' + self._tmp_framework_branch_name])
+
+        # Checkout the default branch
+        self.run_git(['checkout', self.DEFAULT_BRANCH])
+
+        # Create a new branch with checkout -b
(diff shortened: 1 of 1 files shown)


tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,335 @@
...
+from typing import List, Optional, Tuple, Dict
+
+
+class RepoFileMover:
+    """Move files between repos while retaining history."""
+
+    GIT_EXE = 'git'
+    FRAMEWORK_DEFAULT_BASE_BRANCH = 'main'
+
+    def __init__(self, source_repo: str, dest_repo: str,
```
gilles-peskine-arm, line 37: Please be consistent when abbreviating “destination”: either `dst` or `dest`, not both.

tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,335 @@
...
+                        help='Path to the framework repo to move files to')
+    parser.add_argument('--src-branch',
+                        default='move-files-from-mbedtls-to-framework',
+                        required=False,
+                        help='Name of the new branch create in the Mbed TLS repo')
+    parser.add_argument('--dst-branch', default='move-files-into-framework',
+                        required=False,
+                        help='Name of the new branch create in the framework repo')
+    parser.add_argument('--dst-base-branch', default=None, required=False,
+                        help='Base branch in the framework repo to make changes from')
```
gilles-peskine-arm, line 296: That doesn't actually need to be a branch name, it can be any revision. In practice I'd want whatever is the current head of the framework submodule, and I think that should be the default, rather than `main`.

In fact `main` is likely not to work in practice, because it'll typically be some old reference made when the working directory was originally cloned I delete the default branch name in my local clones to avoid keeping that old reference around, so when I didn't pass `--dst-branch-name`, I got the error
```
fatal: 'main' matched multiple (4) remote tracking branches
```

tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,335 @@
...
+
+        # Fetch/checkout the branch that has the moved files on it
+        self.run_git(['fetch', self._source_repo,
+                      self._tmp_framework_branch_name + ':' + self._tmp_framework_branch_name])
+
+        # Checkout the default branch
+        self.run_git(['checkout', self._framework_base_branch_name])
+
+        # Create a new branch with checkout -b
+        self.run_git(['checkout', '-b', self._framework_branch_name])
```
gilles-peskine-arm, line 226: Please add checks for common errors before starting to do anything. At least check that the branches don't already exist, which is likely to be the case the second time the same person uses the script.

tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,335 @@
...
+                dir_files = self.run_git(['ls-files', f]).split()
+                dir_files = [os.path.normpath(dir_file) for dir_file in dir_files]
+                files_to_retain.update(dir_files)
+            else:
+                files_to_retain.add(os.path.normpath(f))
+
+        # Delete all files except the ones we are moving
+        for f in repo_files:
+            if os.path.normpath(f) not in files_to_retain:
+                self.run_git(['rm', f])
```
gilles-peskine-arm, line 142: As far as I understand it (more info in Slack thread), this will in particular remove `framework`. So from that point, when checking out the original head again, the framework submodule is no longer initialized. Then later the framework directory itself disappears, which I don't understand.

I don't fully understand what's going on, but as far as I can tell, you can't have the destination be the `framework` directory of the source, it has to be a separate worktree. This is a very unintuitive limitation that needs to be at least documented.

gilles-peskine-arm (changes requested):
I've done a mostly full code review, more than we usually do in `mbedtls-docs`. Arguably this script could live in the framework repository; for that level of quality, I'd want most of my comments to be handled. For the level of quality in `mbedtls-docs`, all my comments are optional except one: this script seems to mess up if the destination repo is the `framework` subdirectory of the source.

gilles-peskine-arm:
I'm marking this as priority-high because we need this to move files to the framework with the history preserved. Arguably this should be in the framework repository as an essential maintainer tool (even if it'll only be essential for a few months).

tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,335 @@
...
+
+        # Rename files as requested
+        for f in self._file_map:
+            # Git won't let us move things to themselves
+            if self._file_map[f] != f:
+                self.ensure_dirs_exist(self._file_map[f])
+                self.run_git(['mv', f, self._file_map[f]])
+
+        # Commit the result
+        commit_message = "Move some files to framework repository"
```
gilles-peskine-arm, line 152: I would like the ability to specify a commit message. I can do it afterwards by rebasing, and that's not really a problem in the source branch, but in the destination branch, the commit message is behind a merge and git is reluctant to edit history beyond a merge.

gilles-peskine-arm: In fact, if I want to edit the commit message, I need to redo the conflict resolution which is implemented here as `resolve_deletion_conflicts`. That's not easy!

gilles-peskine-arm:
I've now used the script successfully. As far as I can't tell it just can't work if the destination is the `framework` subdirectory of an `mbedtls` worktree. If the source is an `mbedtls` worktree, the script can work, but the source worktree is disrupted because the framework submodule gets deinitialized, and the resulting history may be wrong because any file that is neither tracked nor ignored gets added.

So this script should come at least with a warning, if not enforcement, that both the source and destination should be independent worktrees, freshly checked out, that are not used for doing anything else. Preferably, the script should create these worktrees itself (as `mbedtls-rewrite-branch-style` does, we can reuse some of its code).

mpg:
I've not yet used this script myself, but looking at its result there seems to be extra commits that I wouldn't expect to be present in the history, see https://github.com/Mbed-TLS/mbedtls-framework/pull/55#issuecomment-2404287672

mpg:
> I'm marking this as priority-high because we need this to move files to the framework with the history preserved. Arguably this should be in the framework repository as an essential maintainer tool (even if it'll only be essential for a few months).

Agreed. I've created https://github.com/Mbed-TLS/mbedtls-framework/issues/57 and added it to the first framework epic. (I don't care much about where the script lives, as long as it meets quality requirements.)

gilles-peskine-arm:
> I've not yet used this script myself, but looking at its result there seems to be extra commits that I wouldn't expect to be present in the history

We're preserving the commit history, not just the patch history. So each import branch must have the full history of the project up to the last commit that modifies one of the imported files.

The first time we did this in a pull request, GitHub showed basically all the mbedtls history as part of the pull request, because these were all new commits in the `main` branch of the framework. In each subsequent pull request that imports files into the framework, GitHub just shows recent commits that aren't yet in one of the imports.

There are tools that can rewrite a git branch to only retain commits that modify specific files, and construct a commit history with the same messages and the same patches for this subset of files. But you lose commit IDs (obviously), renamed files, content moved between files, etc. And these tools are harder to use. So I think we've made the right choice.

mpg:
Ah OK, I had misunderstood what the script does.

> There are tools that can rewrite a git branch to only retain commits that modify specific files, and construct a commit history with the same messages and the same patches for this subset of files. But you lose commit IDs (obviously), renamed files, content moved between files, etc. And these tools are harder to use. So I think we've made the right choice.

Yeah, that's what I thought the script was doing, and indeed now that you mention it that wouldn't be the right thing to do. Thanks for clearing that up.

tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,335 @@
...
+        """Get the direct children of a subdirectory (at least the ones
+        that matter to git)"""
+        leaf_files = self.run_git(['ls-files', dir_path]).split()
+        path_len = len(self._path_parts(dir_path)) + 1
+        direct_children = set([
+            os.path.join(
+                *self._path_parts(f)[:path_len]
+            )
+            for f in leaf_files
+        ])
```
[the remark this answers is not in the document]
davidhorstmann-arm: This part removes the suffix that `git ls-files` outputs. We might get:
```
foo/bar/a.c
foo/bar/b.c
foo/baz
```
Which we reduce to just the direct descendants of `foo`, so:
```
foo/bar
foo/baz
```
We never reconstruct the suffix in `_expand_dir_move()`, but we substitute the prefix, e.g. if `foo` is being moved to `new_foo`:
```
new_foo/bar
new_foo/baz
```
This is part of the operations required for the `--except` option.

I was not aware that passing `-C` with a directory had this effect though, that will certainly simplify this.

davidhorstmann-arm:
@mpg I believe I have addressed the main blocker and a few other improvements, so this should now be in a reasonable state to be handed over to someone else for followup.

mpg:
@gilles-peskine-arm Unless you disagree, I'd like to merge this script mostly as it stands now: I think your main comment has been addressed, and in practice we (at least Elena, Valerio and I) have been using this script a few times now without trouble.

Of course that doesn't mean we won't improve it later on if usability or robustness issues start being a problem. But this can be done as a different PR (probably by a different author now that David have moved to other priorities).

Could you give this PR a second review with that in mind?

gilles-peskine-arm:
I'm sorry, I can't promise a review right now. On the other hand, I can see that my one blocker comment has been addressed: the script won't fail in a mysterious way if you try to use it in what is to me the most natural scenario, namely moving files to a submodule of the current checkout. So I've dismissed my review. Since you seem to be happy with it, I'll go ahead and merge.

gilles-peskine-arm:
> I'll go ahead and merge.

Well, not quite: I don't feel comfortable merging it without doing at least a cursory review. I'll try to do that later, but if someone else feels comfortable approving this PR, please go ahead.

tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,395 @@
...
+        try:
+            # Merge in the previously-fetched branch using --allow-unrelated-histories
+            self.run_git(['merge', self._tmp_framework_branch_name, '--allow-unrelated-histories'])
+        except subprocess.CalledProcessError as git_error:
+            # We have conflicts, resolve them if possible
+            if not self.resolve_deletion_conflicts():
+                raise git_error
+
+            # Continue the merge without opening an editor
+            self.run_git(['-c', 'core.editor=/bin/true',
```
Harry-Ramsey, line 292: This fails on MacOS as `/bin/true` does not exist in that directory. Instead it is `/usr/bin/true`.

tools/bin/mbedtls-move-to-framework
```diff
@@ -0,0 +1,395 @@
...
+                          'merge', '--continue'])
+
+        # Edge case: when we are not renaming the files, their deletion in
+        # previous branches moved to the framework will take precedence without
+        # merge conficts. Treat these files specially and restore them.
+        for f in self._file_map:
+            if self._file_map[f] == f:
+                self.run_git(['checkout', self._tmp_framework_branch_name, '--', f])
+                self.run_git(['add', f])
+                self.run_git(['-c', 'core.editor=/bin/true',
```
Harry-Ramsey, line 302: This fails on MacOS as `/bin/true` does not exist in that directory. Instead it is `/usr/bin/true`.

Harry-Ramsey (changes requested):
Just a few changes to make the script work on MacOS.

tools/bin/mbedtls-move-to-framework
gilles-peskine-arm: You can just use `true`, Git will do a path lookup. (Actually you can even write arbitrary shell code there.)

PR #149, opened 2024-07-15
Upgrade dependencies in requirements file

davidhorstmann-arm:
Upgrade Python dependencies to their latest versions in `requirements.txt`

Produced by running:
```
pip-compile --upgrade requirements.in
```


minosgalanakis (approved):
LGTM

PR #153, opened 2024-07-24
There is a wrong include at line 191, in the mentioned exmaple code "…

devprofile98:
this pr fixes the wrong include at line 191.


gilles-peskine-arm (approved):
Thanks!

`net.h` is an old name from PolarSSL. `net_sockets.h` is the new name since Mbed TLS 2.0.

devprofile98:
Nice to know this, Thank you.

PR #154, opened 2024-07-25
PrivateKey inside PrivateKeyInfo should be OCTET STRING

yestyle:
Hi there,

from my testing, the private key generated with openssl has OCTET STRING instead of BIT STRING to contain the RSAPrivateKey.

Also fix representations of some DER structures.


gilles-peskine-arm (approved):
Indeed [PKCS#8](https://datatracker.ietf.org/doc/html/rfc5208#section-5) says that `PrivateKey` is an OCTET STRING. Thanks for the corrections!

PR #155, opened 2024-08-30
Publish security advisories for 3.6.1 and 2.28.9

davidhorstmann-arm:
Publish security advisories approved previously in private.

### DIFF
--- a/security-advisories/mbedtls-security-advisory-2024-08-1.md
+++ b/security-advisories/mbedtls-security-advisory-2024-08-1.md
@@ -1,0 +1,24 @@
+# CTR\_DRBG prioritized over HMAC\_DRBG as the PSA DRBG
+
+**Title** | CTR\_DRBG prioritized over HMAC\_DRBG as the PSA DRBG
+--------- | -----------------------------------------------------
+**CVE** | TODO
+**Date** | TODO
+**Affects** | All versions of Mbed TLS since 2.26.0
+**Severity** | Low
+
+## Vulnerability
+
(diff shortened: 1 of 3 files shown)


security-advisories/mbedtls-security-advisory-2024-08-1.md
```diff
@@ -0,0 +1,24 @@
+# CTR\_DRBG prioritized over HMAC\_DRBG as the PSA DRBG
+
+**Title** | CTR\_DRBG prioritized over HMAC\_DRBG as the PSA DRBG
+--------- | -----------------------------------------------------
+**CVE** | TODO
+**Date** | TODO
```
yanesca, line 6: We might be still waiting for the CVE, but I think we know the date now. (Same for the other two.)

security-advisories/mbedtls-security-advisory-2024-08-1.md
```diff
@@ -0,0 +1,24 @@
+# CTR\_DRBG prioritized over HMAC\_DRBG as the PSA DRBG
+
+**Title** | CTR\_DRBG prioritized over HMAC\_DRBG as the PSA DRBG
+--------- | -----------------------------------------------------
+**CVE** | TODO
```
gilles-peskine-arm, line 5: Please fill in the TODOs

yanesca (dismissed):
I have confirmed that the content is the same as approved in restricted. However, now we know the release date and shouldn't leave that as a TODO. I am not sure if we have the CVEs yet, but if we don't we should raise an issue about it to make sure that we don't forget.

security-advisories/mbedtls-security-advisory-2024-08-1.md
gilles-peskine-arm: The CVE IDs are in the changelog.

davidhorstmann-arm: Apologies, I forgot that there were updates to do. I have filled in both CVE and date.

PR #156, opened 2024-09-04
Add skeleton language server for tests

davidhorstmann-arm:
Create a small python language server that can be used to get information from test `.data` files.

At present this just supports goto-definition from a `.data` file to the corresponding `.function` file.

### DIFF
--- a/tools/bin/mbedtls-test-language-server
+++ b/tools/bin/mbedtls-test-language-server
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+"""Language server for Mbed TLS test files
+"""
+
+# Copyright The Mbed TLS Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+import sys
+from pygls.server import LanguageServer
+from lsprotocol import types
+
+server = LanguageServer('mbedtls-test-server', 'v1')
+
+TEST_FUNCTION = re.compile(r'void (?P<fn_name>[a-zA-Z_][a-zA-Z0-9_]*)\(')
+
+@server.feature(types.TEXT_DOCUMENT_DEFINITION)
+def data_file_goto_definition(ls: LanguageServer, params: types.DefinitionParams):
+    # Get data file
+    data_file = ls.workspace.get_text_document(params.text_document.uri)
+
+    # Get function name under cursor (if there is one)
+    lookup_word = data_file.word_at_position(params.position)
+
+    print(f'Looking for word {lookup_word}', file=sys.stderr)
+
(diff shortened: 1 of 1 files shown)


tools/bin/mbedtls-test-language-server
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+"""Language server for Mbed TLS test files
+"""
```
gilles-peskine-arm, line 3: Can you please share some basic usage instructions? How can users of Vim and VScode make us of this?

tools/bin/mbedtls-test-language-server
```diff
@@ -0,0 +1,67 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+import sys
+from pygls.server import LanguageServer
+from lsprotocol import types
```
gilles-peskine-arm, line 23: It would be nice to have pre-installation instructions. Apparently
```
pip install pygls
```

tools/bin/mbedtls-test-language-server
```diff
@@ -0,0 +1,67 @@
...
+
+server = LanguageServer('mbedtls-test-server', 'v1')
+
+TEST_FUNCTION = re.compile(r'void (?P<fn_name>[a-zA-Z_][a-zA-Z0-9_]*)\(')
```
gilles-peskine-arm, line 27: Technically this also catches auxiliary functions in the header section.

tools/bin/mbedtls-test-language-server
```diff
@@ -0,0 +1,67 @@
...
+
+@server.feature(types.TEXT_DOCUMENT_DEFINITION)
+def data_file_goto_definition(ls: LanguageServer, params: types.DefinitionParams):
+    # Get data file
+    data_file = ls.workspace.get_text_document(params.text_document.uri)
+
+    # Get function name under cursor (if there is one)
+    lookup_word = data_file.word_at_position(params.position)
```
gilles-peskine-arm, line 35: Shouldn't this look for the word at the beginning of the line? And ideally, at the beginning of the word at the beginning of the last non-comment line of the current paragraph. This way it would work anywhere in the test case.

In the equivalent [Emacs code](https://github.com/Mbed-TLS/mbedtls-docs/blob/62f210744f0b510acc21e55c5ed4ac9ad66f292d/tools/emacs/mbedtls-test-data-mode.el#L73), I look for the next thing that looks like a function prototype after a `/* BEGIN_CASE`.

tools/bin/mbedtls-test-language-server
```diff
@@ -0,0 +1,67 @@
...
+
+    print(f'Looking for word {lookup_word}', file=sys.stderr)
+
+    # Substitute '.data' for '.function' in extension to get equivalent function file
+    # TODO how to deal with generated data files?
+    function_file_uri = data_file.uri[:-5] + '.function'
```
gilles-peskine-arm, line 41: Keep everything up to the first dot in the base name.
```suggestion
    function_file_uri = re.sub(r'\.[^\\/]*\Z', r'', data_file.uri) + '.function'
```

(It isn't related to generated data files, by the way.)

gilles-peskine-arm (approved):
Thanks for sharing this! I'm approving this as non-harmful. (I'm not a potential user since I already have this and more [in Emacs](https://github.com/Mbed-TLS/mbedtls-docs/blob/main/tools/emacs/mbedtls-test-data-mode.el).) It's up to you if you want to do my proposed improvements or go and merge.

PR #160, opened 2024-10-21
Change from SVG to PNG for the icon

davidhorstmann-arm:
Browser support for SVGs in img tags is a little spotty.


gilles-peskine-arm (approved):
Thanks!

PR #170, opened 2025-03-06
TLS explainer: ClientHello

gilles-peskine-arm:
Explain TLS records when they're not sent over the network for Wireshark to see.

Explain TLS records, mostly ClientHello for now.

No aim for completeness. I just wrote what I wanted to review https://github.com/Mbed-TLS/mbedtls/pull/9976.

### DIFF
--- a/tools/bin/tls-explainer
+++ b/tools/bin/tls-explainer
@@ -1,0 +1,779 @@
+#!/usr/bin/env python3
+"""TLS message explainer.
+"""
+
+import argparse
+import struct
+import sys
+from typing import List, Tuple
+
+
+class Wire:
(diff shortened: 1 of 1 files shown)


gilles-peskine-arm:
I used this for debugging in a project that is now done. I noticed that some outputs are wrong, so user beware. I didn't have time to debug the code and I'm not planning to actively work on this. I don't think an ad hoc script is the right thing anyway, although I couldn't find something similar when I looked. Wireshark has the right stuff but I don't know how to invoke it on some ad hoc data that is not a network capture.

Given that this is known to be buggy and I don't see much of a future for it, I am withdrawing it for consideration. But feel free to adopt it.

PR #171, opened 2025-03-13
PSA is thread-safe since 3.6.0

gilles-peskine-arm:
[Thanks to Minos for noticing this was out of date.](https://github.com/Mbed-TLS/mbedtls/pull/10004#issuecomment-2721541000)


minosgalanakis (approved):
LGTM

PR #166, opened 2025-02-20
mbedtls-kick-ci: Restore defaulting to the mbedtls repository

gilles-peskine-arm:
This was accidentally broken when adding support for TF-PSA-Crypto.

PR #175, opened 2025-03-26
Fix bug in mbedtls-kick-ci

davidhorstmann-arm:
When a number is supplied with no prefix, we are meant to assume Mbed TLS by default. This we do, but we also remove the first digit of the PR number inadvertantly.

Fix this so that the PR number is interpreted correctly as referring to Mbed TLS and starts the correct CI.


gilles-peskine-arm:
There's a simpler fix in https://github.com/Mbed-TLS/mbedtls-docs/pull/166

PR #176, opened 2025-04-07
Credit reporters for the mbedtls_ssl_set_hostname issue

gilles-peskine-arm:
I should have done that earlier, and actually thought when writing the history section “oh, maybe we should also have credits”, but then forgot to actually go and write them.


davidhorstmann-arm (approved):
LGTM, thanks!

PR #179, opened 2025-05-08
tutorial: Update URL for uIP TCP/IP stack

aaron-sierra:
The current link to https://www.sics.se/... redirects to the https://www.ri.se homepage, where no references to uIP appear in search results.

Update the link to the Wikipedia article about the project, which *may* be more permanent. The article links to the (historical) Github project [adamdunkels/uip](https://github.com/adamdunkels/uip).


gilles-peskine-arm (approved):
Thanks! We probably should suggest something that's more up to date, but incremental improvements are welcome here.

PR #180, opened 2025-05-26
CMake and mbedtls-prepare-build user guide

gilles-peskine-arm:
As requested, here is a description of how the kinds of builds that I typically do as an Mbed TLS maintainer, and how to do them with CMake and with mbedtls-prepare-build.

### DIFF
--- a/kb/development/developer-builds.md
+++ b/kb/development/developer-builds.md
@@ -1,0 +1,727 @@
+# Building Mbed TLS as a contributor
+
+## Introduction
+
+### Purpose of this document
+
+This guide is intended for maintainers and contributors of Mbed TLS and TF-PSA-Crypto. It discusses building the project when working on a contribution and investigating some common problems.
+
+In particular, these projects have a wide range of possible configurations. Therefore a major aspect of this guide is how to work with multiple configurations.
+
+### Developer platform
+
+The primary platform for Mbed TLS and TF-PSA-Crypto is the latest or previous LTS release of Ubuntu Linux on x86_64. Other common platforms are likely to work too, including macOS and WSL, but some of the less common parts of the build system and test scripts may not be well-tested. The project accepts non-disruptive patches to improve the non-Linux experience.
+
+Native Windows is not supported as a primary developer platform. The official build scripts do work on Windows, but you need Perl and Python even for basic usage, and a Unix-style shell for many maintainer scripts.
+
+Some test scripts require specific versions of certain dependencies. See the [CI tooling guide](testing-ci.md#tooling-for-allsh) for more information.
+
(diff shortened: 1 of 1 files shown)


kb/development/developer-builds.md
```diff
@@ -0,0 +1,727 @@
...
+
+```
+mbedtls-prepare-build -p coverage
+```
+
+### Library configuration setup
+
+#### Build in the full configuration
+
+With CMake: ???
```
gilles-peskine-arm, line 222: Pending in https://github.com/Mbed-TLS/mbedtls/pull/10038.

This is something I do very often. It's currently broken with mbedtls-prepare-build in Mbed TLS, because mbedtls-prepare-build needs some refactoring to handle more than one config file. I expect I'll fix that soonish.

kb/development/developer-builds.md
```diff
@@ -0,0 +1,727 @@
...
+
+```
+mbedtls-prepare-build -d build-custom --config-set=MBEDTLS_ECP_RESTARTABLE --config-set=MBEDTLS_ECP_WINDOW_SIZE=2
+```
+
+Limitation: at the time of writing, in Mbed TLS, mbedtls-prepare-build only tweaks options in `mbedtls_config.h`, so this doesn't work for crypto options with Mbed TLS 4.
+
+### Set and unset individual options
+
+CMake doesn't have a direct way to tweak individual config options. You can write `#define` and `#undef` directives in a header file and pass that as the [user config file](#use-a-configuration-file-extension). To run `config.py`, ???
```
gilles-peskine-arm, line 297: A build directory with a tweaked configuration is something I do very often. I don't know how to do it with CMake.

kb/development/developer-builds.md
```diff
@@ -0,0 +1,727 @@
...
+framework/scripts/make_generated_files.py
+```
+
+If build dependencies have changed in the generated files, [run `make dep` in the build tree](#refresh-an-existing-build-directory-after-changing-dependencies).
+
+#### Refresh an existing build directory after changing branches
+
+After changing branches, or rebasing an existing branch, the build scripts may have changed.
+
+With CMake, ???
```
gilles-peskine-arm, line 377: Re-running `cmake ..` in the build directory sometimes works, sometimes not. I'm not sure what to suggest other than removing and re-creating the build directory, but that requires [the ability to reproduce a build directory](https://github.com/Mbed-TLS/mbedtls-docs/pull/180/files#r2107644409).

kb/development/developer-builds.md
```diff
@@ -0,0 +1,727 @@
...
+```
+framework/scripts/make_generated_files.py
+make -C build-foo clean dep
+```
+
+### Builds with drivers
+
+#### Build with all drivers
+
+CMake: ???
```
gilles-peskine-arm, line 399: This is a known missing feature which we'll resolve eventually for the sake of `all.sh`.

PR #182, opened 2025-07-01
Add top-level headers

mpg:
Without them, the list of advisory at [1] looks all wrong because it then picks level-2 headers (Vulnerability, Impact, etc.) instead.

https://mbed-tls.readthedocs.io/en/latest/security-advisories/


minosgalanakis (approved):
LGTM

PR #186, opened 2025-09-05
Upgrade packages in requirements.txt

davidhorstmann-arm:
Upgrade the python packages we use for documentation builds.


yanesca (approved):
LGTM

PR #200, opened 2026-04-02
Updates to security advisories

davidhorstmann-arm:
Update the security advisories. Specifically:

* Add specific affected versions for context serialization
* Add CVE to the title of the SigAlg injection advisory
* Add parentheses for consistency


gilles-peskine-arm (approved):
LGTM

PR #201, opened 2026-04-14
DI: split depends.py

gilles-peskine-arm:
Repository split for depends.py: objectives, design, plan

### DIFF
--- a/architecture/depends-py-split.md
+++ b/architecture/depends-py-split.md
@@ -89,6 +89,9 @@ This leads us to the following file architecture:
 
 ### Export TF-PSA-Crypto knowledge to Mbed TLS
 
+https://github.com/Mbed-TLS/TF-PSA-Crypto/issues/764
+https://github.com/Mbed-TLS/mbedtls/issues/10698
+
 Create a directory where TF-PSA-Crypto can put Python code that can be used by both TF-PSA-Crypto and Mbed TLS. This directory is intended for _project knowledge_, i.e. data that describes TF-PSA-Crypto in a form that Mbed TLS can consume.
 
 We do not put `tf-psa-crypto/scripts` or `tf-psa-crypto/tests/scripts` on the Python import path in Mbed TLS because file names in these directories are likely to clash with file names in the corresponding directories of Mbed TLS, and this could cause confusion as to which module gets loaded.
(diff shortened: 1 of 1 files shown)


architecture/depends-py-split.md
```diff
@@ -0,0 +1,179 @@
...
+### Role of `depends.py`
+
+The script `tests/scripts/depends.py` is used to test the library in a number of different configurations defined by sets of cryptographic mechanisms (or adjacent features, specifically TLS key exchange types). Its goal is to give confidence that the library behaves correctly when only a subset of cryptographic mechanisms is enabled.
+
+The name “depends” comes from the fact that one of the main things that is validated is that library code has correct `#if defined(...)` guards, and tests have correct `depends_on:` guards. However, the testing addresses other risks, such as buffer sizes that depend on the mechanism, and proper handling of requests to use mechanisms that are disabled.
+
+The parts of the library and tests that it is the most relevant to test are:
+
+* Cryptographic mechanisms, to ensure that e.g. any auxiliary function is enabled when it should, and that static buffer sizes are large enough for all enabled mechanisms.
+* Parts of the library that use cryptography in a way that isn't fully generic. This includes many parts of X.509 and TLS because those relate user data with a choice of mechanism.
```
ronald-cron-arm: "those relate user data with a choice of mechanism" I am not sure what this means.

architecture/depends-py-split.md
```diff
@@ -0,0 +1,179 @@
...
+
+The goal of this document is to design and plan the adaptation of the testing done by `depends.py` to the split between TF-PSA-Crypto and Mbed TLS.
+
+## Project overview
+
+### Role of `depends.py`
+
+The script `tests/scripts/depends.py` is used to test the library in a number of different configurations defined by sets of cryptographic mechanisms (or adjacent features, specifically TLS key exchange types). Its goal is to give confidence that the library behaves correctly when only a subset of cryptographic mechanisms is enabled.
+
+The name “depends” comes from the fact that one of the main things that is validated is that library code has correct `#if defined(...)` guards, and tests have correct `depends_on:` guards. However, the testing addresses other risks, such as buffer sizes that depend on the mechanism, and proper handling of requests to use mechanisms that are disabled.
```
ronald-cron-arm: "to use mechanisms that are disabled"
that are disabled by default?

architecture/depends-py-split.md
```diff
@@ -0,0 +1,179 @@
...
+* TF-PSA-Crypto: `tests/scripts/depends.py`: TF-PSA-Crypto entry point, and project-specific data that is not consumed by Mbed TLS.
+* Mbed TLS: `tests/scripts/depends.py`: Mbed TLS entry point and project-specific data, including the X.509/TLS part of reverse dependencies and exclusive groups.
+
+## Actions
+
+### Export TF-PSA-Crypto knowledge to Mbed TLS
+
+Create a directory where TF-PSA-Crypto can put Python code that can be used by both TF-PSA-Crypto and Mbed TLS. This directory is intended for _project knowledge_, i.e. data that describes TF-PSA-Crypto in a form that Mbed TLS can consume.
+
+Code in this directory may call standard Python functions (e.g. set/list/dict functions, `re` functions, …) but should generally not do I/O or import `mbedtls_framework` modules. This way the code is flexible with respect to how it's consumed.
```
gilles-peskine-arm: > should generally not do I/O or import `mbedtls_framework` modules

This was important for the dynamic loading hack in the original version of the outcome analysis split in https://github.com/Mbed-TLS/mbedtls/pull/10676. But now that the directory is an ordinary directory on the load path, and the module is consumed through an ordinary `import` statement, I don;t think it matters anymore.

There are still some potential gotchas, but they're much weaker:

* The code may be consumed by TF-PSA-Crypto or by Mbed TLS, so it must not assume that the project root is TF-PSA-Crypto, and it must not try to load other modules from TF-PSA-Crypto (outside of the same `project_knowledge` directory).
* The code may be loaded with either Mbed TLS's version of the framework or TF-PSA-Crypto's version, so we must be careful when using a recent framework feature.

